PR #142, opened 2025-03-06
Update cosmological inference pipeline

sachaguer:
- Add scripts to postprocess cosmo_val output and get input for cosmoSIS.
- Update config files to account for PSF systematics.


LisaGoh:
Ok! Could you also include the xi_sys_psf.py and tau_from_rho.py scripts that you used to run the cosmosis pipeline?

LisaGoh (approved):

sachaguer:
As said in the readme, they are on my fork of the cosmosis-standard-library and it seems to me that the easiest is to clone it. Any objection with that?

LisaGoh:
Ah fair enough sure!

martinkilbinger (approved):

sachaguer:
@LisaGoh I added m-bias in the ini files. Can you check this is correct?

PR #143, opened 2025-03-06
Convert cosmo_val from script to class

cailmdaley:
This PR turns the various checks performed by `cosmo_val.py` into a set of methods of a `CosmologyValidation` class. This class calculates certain quantities on-demand, reducing memory usage and allowing users to iterate on specific tests without rerunning the full suite of tests each time.  I have also added a runner script `run_cosmo_val.py` that replaces the original script. This should not affect the way results are computed or stored.

I have also added a method to calculate pure EB correlation functions for systematics checks.


LisaGoh (approved):

sachaguer (approved):
One thing that could be added is the computation of the area, effective number of galaxies and stars. After thinking of this I will add it in the shear_psf_leakage script called in cosmo_val.

martinkilbinger (approved):
What an amount of work! This is fantastic :-)

notebooks/cosmo_val/cosmo_val.py
```diff
@@ -1,1236 +1,1383 @@
-# ---
```
martinkilbinger, line 1: At a later stage this file could go to the main directory to be part of the library.

notebooks/cosmo_val/cosmo_val.py
```diff
@@ -1,1236 +1,1383 @@
...
+        theta_min=0.1,
+        theta_max=250,
+        nbins=20,
+        var_method="jackknife",
+        npatch=20,
+        quantile=0.683,
+        theta_min_plot=0.08,
+        theta_max_plot=250,
+        ylim_alpha=[-0.005, 0.05],
+        ylim_xi_sys_ratio=[-0.02, 0.5],
```
martinkilbinger, line 42: To have a cleaner class, we could move those additional parameters to the yaml config file. Can be done later.

martinkilbinger (approved):

PR #144, opened 2025-03-06
Pyproject.toml-based installation, updated dependencies, and automatic Docker builds

cailmdaley:
The shapepipe and sp_validation packages have inconsistent version requirements, causing dependency versions to change when installing both packages in the same environment. This PR attempts to make the dependencies of the two packages consistent (and adds a few new dependencies). 

While I was at it, I switched form the old setup.py installation method to the modern pyproject.toml approach. 

I also plan to set up automatic GitHub builds of a docker container for the package, building on top of the container for the shapepipe package.


pyproject.toml
```diff
@@ -0,0 +1,94 @@
...
+    "Programming Language :: Python :: 3.8",
+    "Programming Language :: Python :: 3.9",
+    "Operating System :: Unix",
+    "Operating System :: MacOS"
+]
+dependencies = [
+    "adjustText",
+    "astropy>=5.0",
+    "colorama",
+    "cosmo-numba @ git+https://github.com/aguinot/cosmo-numba",
```
sfarrens, line 22: This package has an [upper limit on the Python version](https://github.com/aguinot/cosmo-numba/blob/a0e9e0c3910dbf1b23f6c8f0e425efb53529bea4/pyproject.toml#L13). It would be worth checking if we are still locked into using this fork or if we can just use the underlying package. If we need to keep this, you should keep this same versioning limits for `sp_validation` otherwise the installation will break if you use an environemnt with Python >= 3.12.

pyproject.toml
```diff
@@ -0,0 +1,94 @@
...
+    "numpy>=1.14",
+    "opencv-python-headless",
+    "pyccl",
+    "regions",
+    "reproject",
+    "scipy",
+    "seaborn",
+    "shear_psf_leakage==0.2.0",
+    "statsmodels",
+    "treecorr>=5.0",
```
sfarrens, line 49: This appears to conflict with `shear_psf_leakage`. I get the following:

```
sp-validation 0.5.0 depends on treecorr>=5.0
shear-psf-leakage 0.2 depends on treecorr<5.0.0 and >=4.3.3
```

I can see that see that they have the [lower limit set using poetry](https://github.com/martinkilbinger/shear_psf_leakage/blob/6417109c05649769d653d65efdfc1d480d3a1b6e/pyproject.toml#L17) and I guess the upper limit is somehow implicit from the [poetry.lock](https://github.com/martinkilbinger/shear_psf_leakage/blob/develop/poetry.lock) but you might want to look into this.

pyproject.toml
```diff
@@ -0,0 +1,94 @@
...
+docs = [
+    "jupyter==1.0.0",
+    "myst-parser==0.16.1",
+    "nbsphinx==0.8.7",
+    "nbsphinx-link==1.3.0",
+    "numpydoc==1.1.0",
+    "sphinx==4.3.1",
+    "sphinxcontrib-bibtex==2.4.1",
+    "sphinxawesome-theme==3.2.1"
+]
```
sfarrens: This is all fine, but personally I prefer to use `test` for testing dependencies and `dev` or `develop` for all optional dependencies, which can be added with a line like:

```toml
develop = ["sp_validation[test,docs]"]
```

pyproject.toml
```diff
@@ -0,0 +1,94 @@
...
+
+[tool.setuptools]
+packages = ["sp_validation"]
```
sfarrens: It seems that the community standard is moving towards having everything in a `src` directory such that you don't need to specify this.

Dockerfile
```diff
@@ -0,0 +1,5 @@
+# Development image with more bells and whistles
+FROM shapepipe as shapepipe-dev
+
+RUN pip install --no-cache-dir 
+    snakemake
```
sfarrens: I guess you are still working on this file, right?

sfarrens (commented):
Hey @cailmdaley, looking good so far. 👏  I added a few comments on things you can look into. One general point to note is that we probably won't want/need any pinned dependencies in the `pyproject.toml` as we can tag versions of the Docker image instead.

pyproject.toml
cailmdaley: Good point. The cosmo-numba we're using here isn't a fork of anything, so I'll propagate the Python version upper limit from there.

pyproject.toml
cailmdaley: Martin released a new version of shear_psf_leakage that relaxes the treecorr restriction, so this is fixed.

cailmdaley:
thank you for the helpful comments Sam. to your general point, wouldn't it be better to specify python dependencies in the `pyproject.toml` since those dependencies should apply outside the container as well?

cailmdaley:
Getting back to this PR, which has expanded beyond just pyproject.toml/Docker/CI matters to include analysis code changes to `cosmo_val.py`..

@sfarrens I tried to address all of your comments from last time and I'm pretty happy with the non-`cosmo_val.py` changes, so when you're back from vacation I'd welcome a second review!

src/sp_validation/cosmo_val.py
```diff
@@ -1680,21 +2754,21 @@ def hsp_map_logical_or(maps, verbose=False):
...
             valid_pixels = map_comb.valid_pixels
             n_tot = np.sum(valid_pixels)
             n_true = np.count_nonzero(valid_pixels)
             n_false = n_tot - n_true
-            print(f"after map {idx}: frac_true={n_true / n_tot:g}, frac_false={n_false / n_tot:g}")
+            print(
+                f"after map {idx}: frac_true={n_true / n_tot:g}, frac_false={n_false / n_tot:g}"
+            )
 
     return map_comb
```
cailmdaley, line 2151: @sachaguer something funny is going on with the following functions. it looks like they're supposed to be class methods given the `self` argument, but they aren't attached to a class and are probably not evaluated since they're defined inside the above `hsp_map_logical_or` function after the return statement. are these supposed to be members of the `CosmologyValidation` class?

cailmdaley: sorry i didn't preview the right lines, i'm referring to the functions just below `return map_comb`..

cailmdaley: by the way it's possible to have VS Code run `black` on file save, which can make git diffs less polluted when someone else runs `black` later

sachaguer: The pseudo-cl functions are part of the `CosmologyValidation` class. They are on my branch so maybe it's something that went wrong when merging/editing the file.

cailmdaley: ok thanks, i'll fix this to look like your branch. the same thing happened on the develop branch too

cailmdaley:
I think this is ready to be merged pending review, I will do one more run through the whole script to make sure everything works.

README.md
```diff
@@ -33,6 +33,32 @@ tasks that can be performed by `sp_validation` are:
   a shear validation run and performes cosmology inference using the two-point
   correlation function.
 
+## Container Installation (Recommended)
+
+The easiest way to install ShapePipe is via a container. Docker images are automatically built and pushed to the [Github Container Registry (GHCR)](ghcr.io/cosmostat/shapepipe) for each release. This images can be installed and run on most systems (including clusters) with just a few lines of code.
```
martinkilbinger:
```suggestion
The easiest way to install sp_validation is via a container. Docker images are automatically built and pushed to the [Github Container Registry (GHCR)](ghcr.io/cosmostat/sp_validation) for each release. This image can be installed and run on most systems (including clusters) with just a few lines of code.
```

README.md
```diff
@@ -33,6 +33,32 @@ tasks that can be performed by `sp_validation` are:
...
+# open a shell in the container
+apptainer shell --writable sp_validation 
+# and confirm that the installation was successful
+python -c "import sp_validation"
+```
+
+You can also run the image with **Docker**:
+
+```bash
+docker run --rm -it ghcr.io/cosmostat/shapepipe:pyproject_docker python -c "import sp_validation"
```
martinkilbinger:
```suggestion
docker run --rm -it ghcr.io/cosmostat/sp_validation:pyproject_docker python -c "import sp_validation"  
```

martinkilbinger (changes requested):

martinkilbinger (approved):

sfarrens:
Hi @cailmdaley, my apologies for being so slow to finalise this review. I have run the following tests:

1. `pyproject.toml`
- Build on CC-IN2P3 ✅  - build succeeds but I had to mess with the versions of scipy and statsmodels to import the package
- Build on macOS ❌ - build fails

3. `Dockerfile`
- Apptainer build on CC-IN2P3 ✅ - works out of the box
- Local Docker build on macOS ❓- it has not crashed but the build has been running for a very long time, certainly too long for real world users
- Docker pull of pre-built image from ghcr ✅ - no problem pulling, however I would argue that the resulting image is basically unusable because of how slow it runs on Apple Silicon
- I will try one final local build test on my Ubuntu machine tonight just for comparison

In summary, if the scope of this PR is simply to add a `pyproject.toml` setup and a `Dockerfile` that works only for what it absolutely has to work on, then I would say go ahead and merge this PR but just keep in mind that some future PRs will be needed to opened to make this a more usable tool for the community.

cailmdaley:
Thank you Sam! I really appreciate you taking the time to check this. It's good to hear that the Docker build works on another system and a plain pip installation almost works out of the box. 

Yes I think making things work for MacOS is beyond the scope of this PR. Making it usable more widely would be nice longer-term, but for now I think it's most important to get these changes merged so we can use them for the ongoing UNIONS cosmic shear analyses. I will go ahead and merge.

PR #135, opened 2025-01-28
Comprehensive catalogue output

martinkilbinger:
Writing an additional master shear catalogue, with all objects before masking and cuts.
This catalogue includes mask flags (pipeline, SExtractor), the number of exposures, and all
metacalibration information for the user to perform calibration (shear and selection response). 

Check it out
@sachaguer @cailmdaley

PR #145, opened 2025-03-07
V1.5

martinkilbinger:
Classes to calibrate ShapePipe comprehensive catalogues v1.4x and v1.5x.

https://github.com/CosmoStat/sp_validation/pull/135 has been merged, this MR is now good to go.


scripts/create_joint_shape_cat.py
sachaguer, line 1: The final catalogs have their weights labeled 'w_iv' although they correspond to the DES weights. I think the mistake is related to this script but I have not been to pinpoint it yet as I'm not sure how the workflow evolved since the last time I ran this code.

sp_validation/cat.py
sachaguer, line 1: In this file, the default label for the weights given in entry is 'w_iv'. I guess somewhere in the workflow the weights are fed to this function in the 'w' argument hence their wrong label.

notebooks/calibrate_comprehensive_cat.py
sachaguer, line 1: In this file, the function `write_shape_catalog` is called with the DES weights in argument of `w`.  Maybe the `write_shape_catalog` would be clearer if there was a `w_type` argument to specify which type of weights is saved.

sachaguer, line 1: In this file, the function `write_shape_catalog` is called with the DES weights in argument of `w`.  Maybe the `write_shape_catalog` would be clearer if there was a `w_type` argument to specify which type of weights is saved.

sachaguer (commented):
The PR seems good. I did put a few comment to track where the mislabelling of the weights in the output catalog comes from. The easier way is probably to update the write_shape_catalog function with a clearer handle of the weight type.

sachaguer (approved):
Good to go

PR #148, opened 2025-03-14
Update cosmological inference pipeline

sachaguer:
This is again the PR for the cosmological pipeline. The previous one was merged a bit early. This PR contains:
-Update with PSF leakage parameters of the inference pipeline
-Update the ini files to account for m-bias, delta z bias and use intrinsic alignement with a dependance in redshift

The fiducial file we will run with the scale cut should be `cosmosis_pipeline_A_psf.ini`. It seems to be running correctly on my side so just check if anything obvious is missing so that @LisaGoh and @sachaguer run the chains with the same fiducial setup.


cosmo_inference/cosmosis_config/cosmosis_pipeline_A_psf.ini
LisaGoh, line 1: To run it with intrinsic alignment, I believe we also have to include the module [add_intrinsic], see the cosmosis_pipeline_A_ia.ini file

sachaguer: It's updated.

We should have the correct fiducial setup now. Beware just to modify the priors for the psf leakage parameters accordingly before running.

PR #150, opened 2025-03-31
Fix bugs in cosmo_val and add scripts to create glass_mocks

sachaguer:
-Add a folder containing the script used to create GLASS mocks (not specific to UNIONS but applied with UNIONS mask, shape noise and effective density)
-Fixed bugs in cosmo_val.


martinkilbinger (approved):
OK

sachaguer:
Didn't say it in the description but also updated cosmo_val to compute pseudo-cl's as well.

PR #154, opened 2025-05-06
Leakage from auxiliary observables

martinkilbinger:
Computing leakage to galaxy ellipticities from auxiliary (scalar) quantities.
Using shear_psf_leakage.


sachaguer (commented):
The PR looks fine but I have to check what are the conflicts with the main branch.

sachaguer:
Regarding the conflicts:
- `src/sp_validation/cosmology.py`'s conflict is related to loading As and sigma_8. I think the incoming is equivalent to the current change of the PR but is shorter and does not use the `try` `except` block.
-  `src/sp_validation/cosmo_val.py`'s conflict are all missing features in the pseudo-cl calculations that I added in another PR I believe. The incoming changes should be all accepted and the linting reapplied.

martinkilbinger:
Conflicts resolved.

sachaguer (approved):
Ready to merge!

PR #158, opened 2025-07-10
V1.4.6

### LINKED ISSUE #157: Additional flags in v1.[45].X output catalogues
We would like to add: Rii, T_gal un T_PSF

Resolved by #158

martinkilbinger:
Changes for UNIONS catalogue v1.4.6:
- weighted shear response
- cut objects with hlr/hlr_PSF > 0.5**2
- writing additional columns to output shear cat
- fixed bug in mask handling when no mask list argument was given

Resolves #157


sachaguer (approved):
Looks nice. The bounds on the bins for the DES weights are hardcoded. We might try to think of a way to avoid this. (See function get_des_weights in calibration.py)

LisaGoh (approved):

PR #159, opened 2025-07-22
PR catalog-based estimator of pseudo-cls + Notebooks

sachaguer:
This PR contains a few updates of `cosmo_val` I did including the catalog-based estimator for the pseudo-cls.
It also adds notebooks used to compute PTE for pseudo-cls, get the output of the MCMC chains, etc...


martinkilbinger (approved):

sachaguer:
@cailmdaley There are several updates to cosmo_val pending here.

There was some merge conflicts that I tried to solve as best as I can but I think your review is relevant before merging.

cailmdaley (commented):
Initial review of the biggest issues:
  - High – src/sp_validation/cosmo_val.py:55: There’s no comma after
    path_onecovariance=None, so the constructor signature is invalid and the module won’t even import. Nothing in the PR can run until that’s fixed.
  - High – src/sp_validation/cosmo_val.py:2472-2484: When calculate_pseudo_cl_g_ng_cov is called with gaussian_part="OneCovariance", the code still appends gaussian_cov and non_gaussian_cov to the FITS HDU, but those variables were never set in that branch. The first call with that option will raise an UnboundLocalError.
  - High – cosmo_inference/pipeline.sh:41-116: The new pseudo-Cl path uses $root for
    filenames and command arguments, but the script never defines root (the prompts capture xi_root, tau_root, and cosmosis_root). Any inference run now fails immediately with “parameter not set” errors.

cailmdaley:
Some more:
  - High – src/sp_validation/cosmo_val.py:2569: calculate_pseudo_cl_map still calls self.get_pseudo_cls(...), but that helper was renamed to get_pseudo_cls_map. As soon as this branch runs (the default cell_method='map'), the code raises AttributeError and no pseudo-Cl products are generated.
  - High – cosmo_inference/scripts/cosmosis_fitting.py:300: When the user answers “n” to the “USE RHO/TAU_STATS?” prompt, tau_stats_file is set to None, yet the next line calls os.path.exists(tau_stats_file). That produces a TypeError, so the inference pipeline exits before writing the FITS bundle.
  - Medium – cosmo_inference/scripts/cosmosis_fitting.py:291 and :292: The new path concatenations miss the “/” between output_folder and rho_tau_stats, so even existing tau/covariance files look missing and downstream steps fail to pick them up. Should cosmosis_fitting.py keep the old path joins (os.path.join) to avoid similar slip-ups when more branches are added?
  - Medium – src/sp_validation/cosmo_val.py:2573: After the random-noise loop, the code unconditionally deletes e1_rot/e2_rot. If a caller sets nrandom_cell=0 to skip noise realisations, those variables are never defined and the function raises UnboundLocalError.

sachaguer:
All the comments should now be adressed. :smile:

cailmdaley:
Great, all remaining issues I found relate to cosmosis_fitting or masking in cosmo_val; since my PR will also affect these I'd say we can punt on those and you can go ahead and merge!

cailmdaley (approved):

PR #162, opened 2025-10-09
Binned quantities

martinkilbinger:
Compute various catalogue quantities binned in SNR and size, for illustration and testing.


cailmdaley (commented):
## Review: Two critical issues to fix

Thanks for extending the binned quantities functionality! Found two issues that will prevent the notebook from running:

### 1. Stray `s` on line 143 of `leakage_minimal.py`

```python
# Line 143-144
s
quantities, bin_edges = calibration.get_quantities_binned(...)
```

This will raise a `NameError`. Should be removed.

### 2. Undefined variable `which` in `leakage_minimal.py:144`

```python
quantities, bin_edges = calibration.get_quantities_binned(cat_gal, num_bins_x, num_bins_y, which=which)
```

The variable `which` is never defined in the notebook. Probably needs something like:

```python
which = ["number", "response", "leakage", "w_iv", "mag", ...]
```

---

Minor items (non-blocking):
- Duplicate `import numpy as np` in `plot_binned_quantities.py:16`
- Commented-out matplotlib inline block could be cleaned up

Otherwise looks good! 👍

cailmdaley (commented):
## Review: Two critical issues to fix

Thanks for extending the binned quantities functionality! Found two issues that will prevent the notebook from running:

### 1. Stray `s` on line 143 of `leakage_minimal.py`

```python
# Line 143-144
s
quantities, bin_edges = calibration.get_quantities_binned(...)
```

This will raise a `NameError`. Should be removed.

### 2. Undefined variable `which` in `leakage_minimal.py:144`

```python
quantities, bin_edges = calibration.get_quantities_binned(cat_gal, num_bins_x, num_bins_y, which=which)
```

The variable `which` is never defined in the notebook. Probably needs something like:

```python
which = ["number", "response", "leakage", "w_iv", "mag", ...]
```

---

Minor items (non-blocking):
- Duplicate `import numpy as np` in `plot_binned_quantities.py:16`
- Commented-out matplotlib inline block could be cleaned up

Otherwise looks good!

cailmdaley (commented):
## Review: Two critical issues to fix

Thanks for extending the binned quantities functionality! Found two issues that will prevent the notebook from running:

### 1. Stray s on line 143 of leakage_minimal.py

This will raise a NameError. Should be removed.

### 2. Undefined variable which in leakage_minimal.py:144

The variable which is never defined in the notebook. Probably needs to be defined before calling get_quantities_binned().

---

Minor items (non-blocking):
- Duplicate import numpy as np in plot_binned_quantities.py:16
- Commented-out matplotlib inline block could be cleaned up

Otherwise looks good!

cailmdaley (commented):
## Review: Two critical issues to fix

Thanks for extending the binned quantities functionality! Found two issues that will prevent the notebook from running:

### 1. Stray `s` on line 143 of `leakage_minimal.py`

This will raise a `NameError`. Should be removed.

### 2. Undefined variable `which` in `leakage_minimal.py:144`

The variable `which` is never defined in the notebook. Probably needs to be defined before calling `get_quantities_binned()`.

---

Minor items (non-blocking):
- Duplicate `import numpy as np` in `plot_binned_quantities.py:16`
- Commented-out matplotlib inline block could be cleaned up

Otherwise looks good!

martinkilbinger:
Bugs fixed.

cailmdaley (approved):

PR #163, opened 2025-10-09
Binned masks

martinkilbinger:
Compute mask area from binned catalogue, and compare to healsparse where available.


cailmdaley (approved):
Library code changes are very minimal, but looks good to me!

PR #167, opened 2025-11-12
r-band magnitude histograms

martinkilbinger:
Changes
- added scripts to create magnitude histograms
- cleaned up plotting functions


sachaguer (commented):
It looks good to me but similar to PR #154, I have to check where the conflict comes from. It is probably the same cause so resolving conflict in #154 and merging will require to resolve the same conflicts for this PR as well.

sachaguer:
I confirm that the conflicts in `cosmo_val.py` are due to new features in the pseudo-cl computation that are not in this PR. The incoming changes (from the branch develop) should be accepted. 
The conflicts in the `cat_config.yaml` come from a differnt file name for the redshift distribution so the current changes can be accepted.

martinkilbinger:
Fixed the conflicts, I accepted the ones from `develop`.

sachaguer (approved):
Ready to merge

PR #170, opened 2026-01-20
V1.3.6

martinkilbinger:
Running sp_validation on MCCD v1.3.6 catalogue.


cailmdaley:
### Code review

Found 2 issues:

1. **Hardcoded absolute path with username in mask field** - The new SP_v1.3.6 config uses `/home/guerrini/sp_validation/...` which won't work for other users. Other path references in this file use relative paths or the `subdir` pattern.

https://github.com/CosmoStat/sp_validation/blob/35d6fad84e6fac605eba507fdc37e1653bb66a76/notebooks/cosmo_val/cat_config.yaml#L171-L175

2. **Dependencies removed without explanation** - Four dependencies were removed (`cosmo-numba`, `cs_util`, `pymaster`, `shear_psf_leakage`) that are used in multiple files across the codebase. The PR description doesn't mention why these are being removed, which could break existing functionality.

https://github.com/CosmoStat/sp_validation/blob/35d6fad84e6fac605eba507fdc37e1653bb66a76/pyproject.toml#L20-L51

---

Generated with [Claude Code](https://claude.ai/code)

<sub>If this code review was useful, please react with :+1:. Otherwise, react with :-1:.</sub>

martinkilbinger:
I agree, the absolute path is not great, but readable for other users (it works for me).
I think I removed the cosmo-numba dependence, because it requires python 3.12. I am about to create a new docker file with 3.12, we can merge this first to get cosmo-numba back.

cailmdaley:
whoops I was trying out a Claude review command that I didn't know would auto-post the review without allowing me to look at it myself.... sorry. the absolute path is totally fine.

and yes, I see Axel recently updated the minimum Python version for `cosmo_numba`. you said you've already started a new Dockerfile? i'm also happy take care of this, might be easier for me to do because the sp_validation Docker depends on the ShapePipe docker.

martinkilbinger:
I added a numpy upgrade to pass some tests.

cailmdaley (approved):

PR #172, opened 2026-01-26
psf size hack

martinkilbinger:
Implement Fabian's PSF size hack: PSF size is not correctly propagated from shapepipe output to metacal shear catalogues for the `noshear` sample. The hack is to use the `p1` PSF size, since this is the same as the `noshear` case.


cailmdaley:
**Question about weight column in `merge_psf_cat.py`**

https://github.com/CosmoStat/sp_validation/blob/psf_size_hack/scripts/merge_psf_cat.py#L154-L166

This renames the input `w` column to `w_iv` (inverse variance). Should we be using DES weights here instead?

cailmdaley:
hi Martin, are there any changes here that we may want to keep related to docker etc? for example the cfitsio and other pymaster packages?

martinkilbinger:
`libcfitsio-dev` is installed in the ShapePipe Dockerfile, is that sufficient? `pmaster` is in the pyproject.toml.

PR #174, opened 2026-02-02
metacal changes

martinkilbinger:
Wait for https://github.com/CosmoStat/sp_validation/pull/179


src/sp_validation/basic.py
```diff
@@ -564,6 +559,124 @@ def _return():
...
+        - 'flag': boolean mask for flag == 0
+        - 'rel_size': boolean mask for relative size cuts
+        - 'snr': boolean mask for SNR cuts
+        - 'combined': boolean mask for all cuts combined
+
+    """
+    # Handle dictionary input (from metacal class)
+    if isinstance(data, dict):
+        T = data['T']
+        Tpsf = data['Tpsf']
```
fabianhervaspeters, line 616: I think this potentially reintroduces the size-bug as the data is self.ns and the associated Tpsf should be the wrong one of the unsheared branch. If you want to fix it upstream it could be done in _read_data_ngmix. If for some reason data or dict are not instantiated the 'else' below definitely has the problem since the default shear_variant is 'NOSHEAR' andno value is passed when calling the function l.368.

PR #178, opened 2026-02-05
v0.6.0 --- python 3.12

### LINKED ISSUE #126: Docker container for sp_validation
We could easily extend the shapepipe container into a new one that has all the additional sp_validation dependencies. I'm opening this issue to record packages/depedencies that are needed for sp_validation notebooks etc. but aren't listed as dependencies.

- pyccl
- TreeCorr (currently 4.3.3, need more recent version to write GGCorrelation covariance matrices)

On a related note, I noticed that when installing sp_validation in the shapepipe container, the versions of many packages changed which is not ideal for reproducibility. Maybe we can work towards making the version requirements between the two repositories consistent.

martinkilbinger:
- updated to new, python3.12, shapepipe Dockerfile
- pushed to new version 0.6.0
- updated authors and contributors

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -1,0 +1,16 @@
+# Contributors
+
+## Core Authors
+- Martin Kilbinger
+- Cail Daley
+- Sacha Guerrini
+
+## Contributors
+- Emma AyÃoberry
+- Lucie Beaumont
+- Clara Bonini
+- Sam Farrens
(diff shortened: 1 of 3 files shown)


cailmdaley (approved):
Looks good! One small thing: `requires-python` should probably be `>=3.12` rather than `>=3.11`, since cosmo_numba requires 3.12.

PR #180, opened 2026-02-11
Added masking script

LisaGoh:
I have added the masking script that I used to generate the mask, where you can comment out the flags in the `mask` tuple of the `apply_mask()` function, corresponding to the desired catalogue version.

### DIFF
--- a/cosmo_inference/scripts/masking.py
+++ b/cosmo_inference/scripts/masking.py
@@ -3,6 +3,7 @@ import healpy as hp
 import numpy as np
 from multiprocessing import Pool, cpu_count
 import os
+import sys
 # -------------------------
 # Masking logic
 def apply_masks(data, data_ext):
@@ -81,7 +82,7 @@ def build_mask_map_hdf5(filename, nside=512, chunk_size=1_000_000):
 # Example usage
 if __name__ == "__main__":
     filename = "/n17data/UNIONS/WL/v1.4.x/v1.4.5/unions_shapepipe_comprehensive_struc_2024_v1.X.c.hdf5"
-    nside = 8192
+    nside = sys.argv[1]
     out_dir = os.getcwd()
     mask_map = build_mask_map_hdf5(filename, nside=nside, chunk_size=500_000)
     
--- a/cosmo_inference/scripts/masking.py
+++ b/cosmo_inference/scripts/masking.py
@@ -94,5 +94,5 @@ if __name__ == "__main__":
     print('kept area = ', area_obs_deg2)
     
     cl_mask = hp.anafast(mask_map, lmax=3*nside-1)
-    hp.write_map(os.path.join(out_dir, "mask_map_v1.4.5_nside_8192.fits"), mask_map, overwrite=True)
-    np.savez(os.path.join(out_dir, "mask_cls_from_flags_v1.4.5_nside_8192.npz"), ells=np.arange(len(cl_mask)), cl_mask=cl_mask)
+    hp.write_map(os.path.join(out_dir, f"mask_map_v1.4.5_nside_{nside}.fits"), mask_map, overwrite=True)
+    np.savez(os.path.join(out_dir, f"mask_cls_from_flags_v1.4.5_nside_{nside}.npz"), ells=np.arange(len(cl_mask)), cl_mask=cl_mask)
--- a/cosmo_inference/scripts/masking.py
+++ b/cosmo_inference/scripts/masking.py
@@ -4,53 +4,139 @@ import numpy as np
 from multiprocessing import Pool, cpu_count
 import os
 import sys
+from pathlib import Path
+import yaml
 # -------------------------
 # Masking logic
-def apply_masks(data, data_ext):
+
+def apply_condition(array, kind, value): 
     """
-    Returns a boolean mask where True = galaxy passes all cuts (unmasked).
+    Apply a logical condition to a NumPy array and return a boolean mask, based 
+    on the "kind" key in the mask config YAML file.
     """
-    
-    gal_rel_size = np.divide(
+    if kind == "equal":
+        return array == value
+    elif kind == "not_equal":
+        return array != value
+    elif kind == "greater_equal":
+        return array >= value
+    elif kind == "greater":
+        return array > value
+    elif kind == "less_equal":
+        return array <= value
+    elif kind == "less":
+        return array < value
+    elif kind == "range":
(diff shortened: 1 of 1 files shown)


cosmo_inference/scripts/masking.py
```diff
@@ -0,0 +1,98 @@
...
+    npix = hp.nside2npix(nside)
+    pix_area_sr   = 4 * np.pi / npix
+    pix_area_deg2 = (180/np.pi)**2 * pix_area_sr
+    n_obs = mask_map.sum()
+    f_sky_obs = n_obs / npix
+    area_obs_deg2 = n_obs * pix_area_deg2
+    print('kept area = ', area_obs_deg2)
+    
+    cl_mask = hp.anafast(mask_map, lmax=3*nside-1)
+    hp.write_map(os.path.join(out_dir, "mask_map_v1.4.5_nside_8192.fits"), mask_map, overwrite=True)
```
martinkilbinger: use {nside} instead of hardcoded value in filename

martinkilbinger:
Nice, it would be great at some point to consolidate this with the run_joint.py:Mask class. That allows one to get the exact mask combinations for each version via config files, instead of having to change this in the python script (which seems error-prone).

martinkilbinger (changes requested):

cosmo_inference/scripts/masking.py
```diff
@@ -0,0 +1,98 @@
...
+        # (data_ext["1_Faint_star_halos"] == False) &  # v1.4.8
+        # (data_ext["2_Bright_star_halos"] == False) & # v1.4.7
+        (data_ext["4_Stars"] == False) &
+        (data_ext["8_Manual"] == False) &
+        (data_ext["64_r"] == False) &
+        (data_ext["1024_Maximask"] == False) &
+        (data_ext["npoint3"] >= 3) &
+        (gal_rel_size < 3.0) &
+        # (gal_rel_size > 0.707) # v1.4.6/7/8/11
+        (gal_rel_size > 0.5) # v1.4.5
```
cailmdaley: The active cuts combine v1.4.11.2 FLAGS logic (`<= 3`, line 22) with the v1.4.5 size cut (`> 0.5`), but these don't go together — v1.4.11.2 inherits v1.4.6's stricter size cut (`> 0.707`). The filename also says v1.4.5, where FLAGS should be `== 0`. Is this combination intentional, or should the uncommented lines be made consistent for one version?

cosmo_inference/scripts/masking.py
```diff
@@ -0,0 +1,98 @@
...
+    with Pool(cpu_count()) as pool:
+        for pix_indices in pool.imap_unordered(process_chunk, chunks):
+            mask_map[pix_indices] = 1
+
+    return mask_map
+
+# Example usage
+if __name__ == "__main__":
+    filename = "/n17data/UNIONS/WL/v1.4.x/v1.4.5/unions_shapepipe_comprehensive_struc_2024_v1.X.c.hdf5"
+    nside = 8192
```
cailmdaley: We use nside=4096 for the mask throughout the analysis.
```suggestion
    nside = 4096
```

cailmdaley (commented):
Thanks for adding this, Lisa! A few concrete suggestions below — the mechanical fixes are in suggestion blocks you can accept directly.

**Existing cuts in `sp_validation`**: FYI, `src/sp_validation/galaxy.py` has `get_cut_common` and `get_cut_ngmix` implementing very similar quality cuts (FLAGS, IMAFLAGS_ISO, N_EPOCH, magnitude, size). Worth being aware of for cross-referencing.

**Normalized power spectrum**: The `.npz` saves raw `cl_mask` — is there a downstream step that normalizes this into a `.txt`? If so, could that normalization be folded into this script to keep everything in one place?

cosmo_inference/scripts/masking.py
```diff
@@ -0,0 +1,98 @@
...
+    pix_area_sr   = 4 * np.pi / npix
+    pix_area_deg2 = (180/np.pi)**2 * pix_area_sr
+    n_obs = mask_map.sum()
+    f_sky_obs = n_obs / npix
+    area_obs_deg2 = n_obs * pix_area_deg2
+    print('kept area = ', area_obs_deg2)
+    
+    cl_mask = hp.anafast(mask_map, lmax=3*nside-1)
+    hp.write_map(os.path.join(out_dir, "mask_map_v1.4.5_nside_8192.fits"), mask_map, overwrite=True)
+    np.savez(os.path.join(out_dir, "mask_cls_from_flags_v1.4.5_nside_8192.npz"), ells=np.arange(len(cl_mask)), cl_mask=cl_mask)
```
cailmdaley: Use the `nside` variable in filenames (also addresses Martin's comment). Added trailing newline.
```suggestion
    hp.write_map(os.path.join(out_dir, f"mask_map_v1.4.5_nside_{nside}.fits"), mask_map, overwrite=True)
    np.savez(os.path.join(out_dir, f"mask_cls_from_flags_v1.4.5_nside_{nside}.npz"), ells=np.arange(len(cl_mask)), cl_mask=cl_mask)
```

cosmo_inference/scripts/masking.py
LisaGoh: I modified it to be a command line input instead!

cosmo_inference/scripts/masking.py
LisaGoh: Right these have been arbitrarily commented out, I will try to integrate it with the Mask class as pointed out by Martin so that it automatically reads in the options from the config file!

LisaGoh: I've refactored the code to take the catalogue version as a command line input, and read ni the mask options from the config YAML file instead, so it's much cleaner now!

LisaGoh:
I've also added the bit where I normalise the Cl's by the area

martinkilbinger (approved):

cosmo_inference/scripts/masking.py
```diff
@@ -0,0 +1,246 @@
...
+import numpy as np
+from multiprocessing import Pool, cpu_count
+import os
+import sys
+from pathlib import Path
+import yaml
+# -------------------------
+# Masking logic
+
+def apply_condition(array, kind, value):
```
martinkilbinger, line 12: This is fine, but it would be more robust to use the already implemented function Mask().apply; e.g. if in the future more conditions are added, this will break.

cailmdaley:
Thank you Lisa this looks great!

PR #185, opened 2026-03-26
Updated config space paper plots

LisaGoh:
Hello, I cleaned up and updated all the notebooks to create the plots in the configuration space paper!


cailmdaley:
nice! about to review but i wanted to ask if you'd be interested in doing something like what was done in the SPT-3G D1 paper on the diagonal? both pretty and more information, will make it more likely that people use our plots. it's implemented in Karim's [fork of getdist](https://github.com/benabed/getdist/tree/upper_triangle_whisker) (example notebook [here](https://github.com/benabed/some_jupyter/blob/main/wriangle_plot_example.ipynb). happy to prototype it in order to not make more work for you

Screenshot 2026-03-27 at 19 01 20

cailmdaley:
Looks great Lisa, nice cleanup! Two small things I noticed:

**1. Whisker plot separator loop** (`S8_om_sigma8_whisker.ipynb`) — the `for ax in axs:` loop on the separation lines doesn't use `ax`; it draws on `ax1`/`ax2`/`ax3` directly, so each line gets drawn 3× redundantly. Doesn't affect the output, just a no-op loop.

**2. `if` → `elif` in `get_chi2.ipynb`** — the chain-loading cell has:
```python
if 'nautilus' in root:
    samples = ...
if 'mh' in root:       # ← should be elif
    ...
else:
    samples = ...       # overwrites nautilus case
```
For a nautilus root, the `else` branch (polychord format) runs after the nautilus branch, overwriting it. `contours.ipynb` has this correct with `elif`. Dormant since the getdist files already exist on disk, but would bite on a fresh rerun.

LisaGoh:
Thank you! Okie will look into adding the separate 1D posteriors

cailmdaley (approved):
Looks good, thanks Lisa!

PR #186, opened 2026-04-20
refactor: dedupe FootprintPlotter, import from cs_util

cailmdaley:
## Summary

`FootprintPlotter` was duplicated between `sp_validation.plots` and `cs_util.plots`, with the local copy stale relative to the canonical cs_util version. This PR deletes the local copy and **imports cs_util's `FootprintPlotter` directly at every call site** — the elegant end state, as if the class had always lived in cs_util.

## What changed

- **`plots.py`** — the local `FootprintPlotter` class is gone; the module no longer references it at all (no re-export shim).
- **`cosmo_val.py`** — drops `from .plots import FootprintPlotter`; uses the module's existing `cs_plots` alias → `cs_plots.FootprintPlotter()`.
- **`notebooks/cosmo_val/catalog_paper_plot/plot_gaia_sky.py`** and **`notebooks/demo_create_footprint_mask.py`** — import `from cs_util.plots import FootprintPlotter` directly.
- **`pyproject.toml`** — `cs_util>=0.1.5` → `>=0.1.9` (guarantees the import resolves).

### ⚠️ Intended breaking change

`from sp_validation.plots import FootprintPlotter` no longer resolves. This is deliberate — cs_util is the home for the class, and every caller in the repo now imports it from there. No shim is kept.

## Caller audit

cs_util 0.1.9's `FootprintPlotter` is a superset of the old local API (same `__init__(nside_coverage=32, nside_map=2048)` and public methods, plus `vmin`/Milky-Way/extent improvements). All call sites were inspected and are compatible:

- `cosmo_val.py:plot_footprints` — `FootprintPlotter()` + `create_hsp_map` + `plot_region`
- `plot_gaia_sky.py` — `FootprintPlotter(nside_coverage=32, nside_map=2048)` + `plot_area(..., vmax=...)`
- `demo_create_footprint_mask.py` — `FootprintPlotter()` + `plot_region` / `plot_footprint_as_hp`

## Tests now run in CI

Previously **no** unit tests ran in CI: the only live workflow (the docker build) did `python -c "import sp_validation"`, and `ci-build.yml` was dead (triggered on `master`, Python 3.8, conda, `setup.py test`).

- **`tests/test_plots.py`** (new) — guards the dedup contract (cs_util is the home; `sp_validation.plots`/`cosmo_val` don't re-expose it) and exercises `create_hsp_map` on a tiny input.
- **`testpaths`** fixed from `["sp_validation"]` (matched nothing under the src layout, so pytest recursed the whole tree and choked on a stray `*test*` script) → `["src/sp_validation/tests"]`.
- **`@pytest.mark.slow`** on the catalog-data-dependent tests (additive-bias reads the real SP_v1.4.5 catalog; `catalog_paths_exist` checks files on disk), restoring a fast hermetic `-m "not slow"` suite (**2:54 → 7s**); `catalog_paths_exist` also skips when the data isn't mounted.
- **Dockerfile** installs the `[test]` extra; **`deploy-image.yml`** runs `pytest -m "not slow"` inside the built image (which carries the full system-dependency stack), so tests run on every push.
- Deleted the dead **`ci-build.yml`**.

— Claude on behalf of Cail


cailmdaley:
Live smoke test passed inside the shapepipe container (Python 3.12, cs_util 0.1.9):

```
from sp_validation.plots import FootprintPlotter
import cs_util.plots
fp = FootprintPlotter()
# FootprintPlotter.__module__ -> cs_util.plots
# FootprintPlotter is cs_util.plots.FootprintPlotter -> True
# list(fp._regions.keys()) -> ['NGC', 'SGC', 'fullsky']
# all('vmin' in r for r in fp._regions.values()) -> True
```

Re-export is transparent; same class object is returned, `_regions` matches cs_util's (with `vmin`). Ready for review.

martinkilbinger (approved):

PR #206, opened 2026-06-20
chore: ruff format + region-aware lint gate

cailmdaley:
## What this does, and why

Wires `ruff` (one fast tool that both formats and lints) into `sp_validation` under a **warn-locally / account-on-`develop`** discipline: lint never interrupts local flow, but the shared branch stays clean and every regression is pinned to a person.

After this lands: every commit auto-applies ruff's safe fixes and warns on the rest; the library is formatted and lint-clean; and `develop` is gated — a red check blocks the merge while you're told exactly what to fix.

## Two postures

**Locally (pre-commit) — auto-fix the safe stuff, warn on the rest.** `ruff format` and `ruff check --fix`'s *safe* fixes (formatting, import sorting, unused imports) auto-apply — a one-time re-stage, not a block. What ruff won't safely fix — undefined names (F821), unused variables (F841) — prints as a **warning** and never blocks the commit. (That split is ruff's own safe/unsafe-fix taxonomy, not a hand-curated list.) Bloat guards (`nbstripout`, large-files) still block.

**On `develop` (CI) — the gate.** On push to `develop` and PRs into it, CI runs the full region-aware policy. On failure it goes **red (blocks the merge)** *and* tells the author where they already are:
- **on a PR** → an auto-updating **comment on the PR** with the full violation list, which turns green once ruff passes;
- **on a direct push** to `develop` (no PR) → a per-committer **auto-closing lint-debt issue**, @-mentioning + assigning them.

Fork PRs are covered via `pull_request_target`, hardened with `uvx --no-config` so an untrusted PR can't redirect ruff's download to a trojaned index.

## Why the policy is region-aware

A library and a one-off analysis script don't deserve the same standard. `src/sp_validation` is strict — zero ignores. The peripheral trees (`workflow/`, `papers/*/scripts`, the per-paper analysis scripts, `scripts/`, `scratch/`, `cosmo_inference/`) are formatted, safe-autofixed, and have their real bugs fixed, but waive a few structurally-intentional patterns: **E402** (imports after `sys.path`/backend setup) and **F403/F405** (intentional star-imports in interactive scripts). **E501** is waived everywhere — the formatter owns line width (ruff's own guidance). Genuine bug-classes stay enforced everywhere.

## The commits

1. `style(ruff)` — repo-wide format pass (mechanical; listed in `.git-blame-ignore-revs`).
2. `chore(ruff)` — region-aware lint policy in `pyproject.toml`.
[...]

PR #207, opened 2026-06-20
refactor(cosmo_val): god-class → tested, primitive-backed package

### LINKED ISSUE #203: Decompose run_cosmo_val into a modular workflow (split the cosmo_val god class)
Part of the sp_validation restructuring (#202). `run_cosmo_val` has grown into a god function that does too much in one place. Break it into a modular Snakemake workflow with clear, independently-runnable stages writing to `results/`, so more than one person can run and extend the validation.

— Claude on behalf of Cail

cailmdaley:
## What

Turns the 3,700-line `CosmologyValidation` god-class into a clean, tested `cosmo_val/` package backed by shared primitive modules. Behavior-preserving throughout, except **three clearly-committed bug fixes** (below). Built in reviewable layers — review per commit or per section:

1. **Package split.** `cosmo_val.py` → a `cosmo_val/` mixin package, one module per diagnostic (`pseudo_cl`, `psf_systematics`, `real_space`, `catalog_characterization`, `pure_eb`, `cosebis`, `core`). `CosmologyValidation` composes the mixins; every call site is unchanged. The 85 methods are conserved exactly.
2. **Cleanup.** Per-module tightening plus cross-module dedup (~−150 lines): `_output_path`/`_binning`/`_cprint` helpers, `get_cov_from_onecov` folded into `statistics.cov_from_one_covariance`, `chi2_and_pte` standardized to `solve`+`sf`.
3. **Thinning.** Stateless computation pulled out of the orchestrator into shared primitives: the survey-stat family → `survey.py`, and the NaMaster estimator → a **new top-level `pseudo_cl.py`**. `glass_mock`'s drifted estimator copies now route through the shared primitives.
4. **New test coverage.** `test_pseudo_cl.py` adds the first numeric pins for the pseudo-Cℓ estimator (a Paper II deliverable that had none) — golden values for the three binning schemes, the occupancy map, and the map/catalog EE·EB·BB spectra.

## The architecture this lands

`cosmo_val/` is the stateful **orchestrator** (holds config/paths/results, loads catalogs, sequences diagnostics, caches, plots, saves). The top-level modules (`b_modes`, `rho_tau`, `cosmology`, `statistics`, `survey`, `pseudo_cl`) are the stateless **primitives** that analysis scripts also import directly — which is why they live outside the class.

## Bugs fixed (each its own commit)

- **`demo_create_footprint_mask`** called `cosmo_val.hsp_map_logical_or`, which lives in `plots.py` — repointed.
- **Aperture-mass DES calibration** used scalar `R` for every version while 2pcf used the catalog-averaged `R11/R22`; unified so DES is calibrated the same way in both.
- **Pseudo-Cℓ noise-debiasing was non-reproducible** — `apply_random_rotation` called `np.random.seed()` with no argument, re-seeding from entropy every call. Now threads a generator seeded from a new `CosmologyValidation.cell_seed` (default 8192): same Monte-Carlo scheme, reproducible run-to-run.

## Verification

`ruff check src/sp_validation` clean; all modules import (the relative-import depth is exercised, not just linted). Value-drift (`test_cosmo_val`, `test_b_modes`) green; the new `test_pseudo_cl` pins pass; full non-slow suite: 136 passed.

Stacked on #206. Draft for review.

— Claude on behalf of Cail


cailmdaley:
@sachaguer you may want to take a look at the pseudo_cl random seed behavior and where to integrate in the namaster utils, although the latter doesn't have to be part of this PR!

sachaguer (commented):
I checked the seeding for the pseudo_cl and it looks fine to me.
Some of the namaster_utils are in pseudo_cl.py. Functions could still be broken into more elementary pieces but it is good as is and can be kept for a future PR.

PR #219, opened 2026-07-01
glass: extension of the existing script to the tomographic case

### LINKED ISSUE #213: Tomographic GLASS mock generation
**Goal** 

Upgrade the existing code to generate GLASS mocks with tomography. This is a high priority step as being able to generate UNIONS-like mock will allow to test all the downstream sp_valiation tasks such as data vector measurements with tomography, the conversion to CosmoSIS-like data vectors, and inference.

**To do**

The task is relatively elementary as GLASS already allows to handle tomography. What needs to be implemented is a nice API/config to provide the inputs to the GLASS mocks and tests to the scripts.

**Input**

- Redshift distribution 🟢  
- Number density and shape noise for each redshift bin 🟢 
- A mask 🟢 
- Cosmology 👍

sachaguer:
This PR is corresponding to the issue #213.

Current progress allows me to run a tomographic GLASS mock. I am still in the validation process.

The API has been revised to work with config files and allow more flexibility on the parameters chosen before running the simulation.


config/glass_mock/config_glass_mock.yaml
```diff
@@ -26,8 +26,8 @@ dx: 120.0
 zmax: 3.0
 
 # --- galaxy population (downstream of map generation) ---
-nbins: 1
-n_arcmin2: 6.0905
+nbins: 6
+n_arcmin2: [1, 1.02, 1.05, 0.85, 0.9, 1.2]
 sigma_e: 0.2684
```
martinkilbinger, line 31: should we implement sigma_e per tomo bin?

config/glass_mock/config_glass_mock_test.yaml
```diff
@@ -36,6 +36,6 @@ phz_sigma_0: 0.03
 # --- Runtime options ---
 limber: True
 mask_path: /n09data/guerrini/glass_mock_v1.4.6_rerun/mask_nside4096.fits
-nz_path: /home/guerrini/sp_validation_cosmostat/config/glass_mock/redshift_distribution_tomo.txt
+nz_path: /home/guerrini/sp_validation_cosmostat/config/glass_mock/test_data/redshift_distribution_tomo.txt
```
martinkilbinger, line 40: would be good to have this point to a neutral place, e.g. `/n17data/UNIONS/WL/v2.0/...`

config/glass_mock/config_glass_mock.yaml
martinkilbinger: Ah I see in the py script usage that this can be a scalar or array. Good. Maybe add a comment here about the two syntax options.

martinkilbinger (changes requested):

config/glass_mock/config_glass_mock.yaml
sachaguer: > should we implement sigma_e per tomo bin?

Exactly, it can be both. I will add a comment.

config/glass_mock/config_glass_mock_test.yaml
sachaguer: The path should lead to the example data pushed to the repo for demonstration purpose. What I suggest is to keep it that way while I am testing but once done erase the user specific start of the path name and add a placeholder like `<path_to_sp_validation>`

martinkilbinger (approved):

scripts/glass_mock/make_glass_sim.py
```diff
@@ -0,0 +1,416 @@
...
+            if not os.path.exists("config/glass_mock/test_data/"):
+                raise FileNotFoundError(
+                    "The test must be run from the root of the repository."
+                )
+            if not os.path.exists("config/glass_mock/test_data/mask.fits"):
+                raise FileNotFoundError(
+                    "Check that you downloaded the test mask from the repository."
+                )
+            if not os.path.exists(
+                "config/glass_mock/test_data/redshift_distribution.txt"
```
cailmdaley, line 87: `--test` mode is broken as committed: this file doesn't exist — the PR ships `redshift_distribution_non_tomo.txt` and `redshift_distribution_tomo.txt`, so this raises `FileNotFoundError` immediately. Also `config_glass_mock_test.yaml` isn't used by `--test` at all (this branch builds config via `from_planck18` directly) — the two test paths have diverged; simplest fix might be to make `--test` just load the test yaml.

config/glass_mock/config_glass_mock.yaml
```diff
@@ -0,0 +1,41 @@
...
+nbins: 6
+n_arcmin2: [1, 1.02, 1.05, 0.85, 0.9, 1.2]
+sigma_e: 0.2684
+bias: 1.2
+phz_sigma_0: 0.03
+#ia_bias: null
+
+# --- Runtime options ---
+limber: True
+mask_path: /n09data/guerrini/glass_mock_v1.4.6_rerun/mask_nside4096.fits
```
cailmdaley, line 38: Still the hardcoded personal paths from Martin's earlier round — the committed `config/glass_mock/test_data/` assets in this PR seem like the natural replacement (at least for the test config).

src/sp_validation/glass_mock.py
```diff
@@ -110,6 +131,34 @@ def Oc(self) -> float:
         """Cold dark matter density (CDM = matter - baryons), pre-neutrino."""
         return self.Om - self.Ob
 
+    def check_consistency(self):
```
cailmdaley, line 134: `from_yaml`, `check_consistency`, and the `shape_noise`/`number_density` helpers are pure Python (no GLASS import needed) — a few unit tests (list-length mismatch → ValueError, per-bin n(z) normalization, scalar-vs-list broadcasting) would run in today's CI and lock in the config semantics cheaply.

cailmdaley (commented):
Reviewing as the draft it is — the shape is good (list-or-scalar per-bin config, `TOM_BIN_ID` tagging, per-bin validation helpers), and most of what's left is finishing touches. Inline below; two headlines:

**The dependency change is verified good news.** We empirically checked `glass[examples]==2026.2` against every glass/cosmology symbol this branch uses: clean install (glass-ext-camb 2023.6 and cosmology-compat-camb ride along via the `examples` extra), all APIs present with matching signatures, and the previously-broken map-generation path (`build_shells` → `matter_cls` → `generate` chain) now runs end-to-end, seed-deterministic. So this supersedes the old three-way pin — please also drop the now-stale `xfail` on `test_matter_maps_are_seed_deterministic` (its reason cites `Cosmology.from_camb`, which this branch no longer calls, and the test's call chain passes under 2026.2) and replace the commented-out pin block in pyproject.toml with a one-line note that 2026.2 resolves the whole stack.

**Cross-PR wiring** (also noted on #221): #221 reads the tomo bin column from `cc[ver]["shear"]["tomo_bin_ids"]`; this PR writes the literal column `TOM_BIN_ID`. Neither PR adds the config entry connecting them — worth wiring so the GLASS→pseudo-Cℓ validation loop actually closes, ideally as a small end-to-end 2-bin test.

— Fable, on behalf of Cail

scripts/glass_mock/make_glass_sim.py
sachaguer: This will be cleaned when my whole coherent testing will be fulfilled.

config/glass_mock/config_glass_mock.yaml
sachaguer: Same as above

sachaguer:
I would tend to say that the results shown in #240 is sufficient to merge this PR. We could reopen one if any bug is spotted later. I will clean the test configs before pushing the button but let me know if you have any final comment.

cailmdaley:
sounds good to me

PR #221, opened 2026-07-03
Tomographic pseudo-cl

### LINKED ISSUE #214: Tomographic pseudo-cls
**Goal**

Add functions to `cosmo_val` to perform the computation of pseudo-cls data vector and covariances in the tomographic case.

**To do**

- Some of the implementation already exists in Euclid's FDQA.
- It can be a good time to break the harmonic space related computation in elementary functions.
- Add functions for the covariance based on OneCovariance.
- See how to incorporate Gaussian simulations. This is quite heavy and MPI paralelisation worked great. Relying on MPI makes the incorporation to `cosmo_val` more complex. It maybe is better to have it as an external script.

sachaguer:
I upgraded the harmonic space code to compute tomographic data vectors.

I modified the API related to any utility function in pseudo_cl.py. That way, any user of the repo can call the same "elementary" functions of `cosmo_val` in notebooks or scripts outside of `cosmo_val`. This summarises two separate behaviour:
- Being able to read and interpret fits files containing the catalog. This is carried out in ``CosmologyValidation``.
- Computing relevant quantities from well-defined input, e.g. ellipticity, weight and not a catalog. This is done in the `pseudo_cl.py` script.

The current update passes CI after some tweaks to tests that were not compliant with the new API. Extra tests to validate the tomography approach would be valuable to write at some point.

The PR is not ready, I will keep working to have a similar thing for the covariance. But the current state of the PR allows to start validating the implementation of the tomographic GLASS simulations and the upgraded cosmo_val.

### DIFF
--- a/src/sp_validation/cosmo_val/core.py
+++ b/src/sp_validation/cosmo_val/core.py
@@ -1,5 +1,6 @@
 # %%
 import copy
+import itertools
 import os
 import re
 from pathlib import Path
@@ -7,6 +8,7 @@ from pathlib import Path
 import colorama
 import numpy as np
 import yaml
+from astropy.io import fits
 from shear_psf_leakage import run_object, run_scale
 
 from ..b_modes import (
@@ -96,6 +98,10 @@ class CosmologyValidation(
         noise debiasing, making those realizations reproducible run-to-run.
     cosmo_params : dict, optional
         Cosmological parameters to pass to get_cosmo(). If None, uses Planck 2018.
+    compute_tomography : bool, default False
+        Whether to compute tomographic correlation functions and pseudo-C_ell.
+    force_run : bool, default False
+        If True, forces re-computation of results even if cached outputs exist.
 
     Attributes
     ----------
@@ -222,6 +228,8 @@ class CosmologyValidation(
         path_onecovariance=None,
         cosmo_params=None,
         blind=None,
+        compute_tomography=False,
+        force_run=False,
     ):
         self.rho_tau_method = rho_tau_method
         self.cov_estimate_method = cov_estimate_method
@@ -251,6 +259,8 @@ class CosmologyValidation(
         self.nside_mask = nside_mask
         self.path_onecovariance = path_onecovariance
         self.blind = blind
+        self.compute_tomography = compute_tomography
+        self.force_run = force_run
 
         assert self.cell_method in ["map", "catalog"], (
             "cell_method must be 'map' or 'catalog'"
@@ -631,3 +641,36 @@ class CosmologyValidation(
         print()
 
         return summary
+
+    def _get_tomo_bins(self, version):
+        """
+        Return the tomo_bin_ids for a given version. If the version does not have tomography, return None.
+
+        Returns
+        -------
+        tomo_bin_ids : list or None
+            List of unique tomographic bin IDs for the version, or None if no tomography is available
+        tomo_bin_pairs : list of tuples or None
+            List of unique pairs of tomographic bin IDs (including self-pairs) for the version, or None if no tomography is available
+        """
+        if "tomo_bin_ids" in self.cc[version]["shear"]:
+            self.print_cyan(
+                f"Extracting tomography information from version {version}."
+            )
+            cat_gal = fits.getdata(self.cc[version]["shear"]["path"])
+            tomo_bin = cat_gal[self.cc[version]["shear"]["tomo_bin_ids"]]
+            tomo_bin_ids = np.unique(tomo_bin)
+            tomo_bin_ids = tomo_bin_ids[
+                tomo_bin_ids > 0
+            ]  # Exclude zero or negative bins
+            self.print_cyan(
+                f"Found {len(tomo_bin_ids)} tomographic bins for version {version}: {tomo_bin_ids}."
+            )
+
+            tomo_bin_pairs = list(
+                itertools.combinations_with_replacement(tomo_bin_ids, 2)
+            )
+            return tomo_bin_ids, tomo_bin_pairs
+        else:
+            self.print_cyan(f"Version {version} does not have tomography information.")
+            return None, None
--- a/src/sp_validation/cosmo_val/core.py
+++ b/src/sp_validation/cosmo_val/core.py
@@ -656,12 +656,12 @@ class CosmologyValidation(
         tomo_bin_pairs : list of tuples or None
             List of unique pairs of tomographic bin IDs (including self-pairs) for the version, or None if no tomography is available
         """
-        if "tomo_bin_ids" in self.cc[version]["shear"]:
+        if "tomo_bin_col" in self.cc[version]["shear"]:
             self.print_cyan(
                 f"Extracting tomography information from version {version}."
             )
             cat_gal = fits.getdata(self.cc[version]["shear"]["path"])
-            tomo_bin = cat_gal[self.cc[version]["shear"]["tomo_bin_ids"]]
+            tomo_bin = cat_gal[self.cc[version]["shear"]["tomo_bin_col"]]
             tomo_bin_ids = np.unique(tomo_bin)
             tomo_bin_ids = tomo_bin_ids[
                 tomo_bin_ids > 0
--- a/src/sp_validation/cosmo_val/core.py
+++ b/src/sp_validation/cosmo_val/core.py
@@ -219,7 +219,7 @@ class CosmologyValidation(
         power=1 / 2,
         n_ell_bins=32,
         ell_step=10,
-        pol_factor=True,
+        pol_factor=-1,
         cell_method="map",
         noise_bias_method="analytic",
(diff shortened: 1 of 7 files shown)


src/sp_validation/tests/test_pseudo_cl.py
```diff
@@ -491,16 +513,16 @@ def test_apply_random_rotation_reproducible_with_seed(cv, cat_and_params):
...
     npt.assert_array_equal(a1, b1)
     npt.assert_array_equal(a2, b2)
 
-    c1, _ = cv.apply_random_rotation(e1, e2, np.random.default_rng(7))
+    c1, _ = apply_random_rotation(e1, e2, np.random.default_rng(7))
     assert not np.allclose(a1, c1)
 
-    d1, _ = cv.apply_random_rotation(e1, e2)
-    f1, _ = cv.apply_random_rotation(e1, e2)
+    d1, _ = apply_random_rotation(e1, e2, None)
```
martinkilbinger: `None` is not needed, since this is the default arg

martinkilbinger (changes requested):
Looks good, although there is some code duplication between pseudo_cl and pseudo_cl_tomo. Could we just have one function that internally checks whether tomography or not needs to be computed? Or set zbin=1 for non-tomography and have the same structure for all cases (tomography but in some cases with one bin)?

sachaguer:
> Looks good, although there is some code duplication between pseudo_cl and pseudo_cl_tomo. Could we just have one function that internally checks whether tomography or not needs to be computed? Or set zbin=1 for non-tomography and have the same structure for all cases (tomography but in some cases with one bin)?

Done. I merged both functions with an API where the `calculate_pseudo_cl` function takes an argument `compute_tomography`. Non-tomography dictionary keys are now "tomo_bin_all_tomo_bin_all".

This is not optimal because columns are duplicated. `tomo_bin_a==tomo_bin_b` statements could be put a bit everywhere to cope with this issue.

src/sp_validation/cosmo_val/pseudo_cl.py
```diff
@@ -451,209 +449,359 @@ def calculate_pseudo_cl_g_ng_cov(self, gaussian_part="iNKA"):
...
+        params,
+        nside,
+        cat_gal,
+        n_gal_map,
+        unique_pix=None,
+        idx=None,
+        idx_rep=None,
+        wsp=None,
+    ):
+        cl_noise = np.zeros_like()
```
cailmdaley: `np.zeros_like()` needs an argument — this raises `TypeError` on any auto-bin pair, including the plain non-tomographic `("all","all")` case. Presumably `np.zeros_like(cl_shear)` as before the refactor.

src/sp_validation/cosmo_val/pseudo_cl.py
```diff
@@ -451,209 +449,359 @@ def calculate_pseudo_cl_g_ng_cov(self, gaussian_part="iNKA"):
...
+            f"Computing pseudo-Cl's for tomographic bins {tomo_bin_a} and {tomo_bin_b}..."
+        )
 
-        ell_eff, cl_shear, wsp = self.get_pseudo_cls_map(shear_map, n_gal_map)
+        # Load data and create shear and noise maps
+        cat_gal = fits.getdata(self.cc[ver]["shear"]["path"])
 
-        cl_noise = np.zeros_like(cl_shear)
-        rng = np.random.default_rng(self.cell_seed)
+        if tomo_bin_a != "all" and tomo_bin_b != "all":
```
cailmdaley: These branches are swapped: real bin pairs (`!= "all"`) take the *unmasked* full catalog, so every tomographic auto/cross spectrum computes the same full-catalog Cℓ; and the `"all"` case falls into the mask branch, where `tomo_bin_id == "all"` on an int column is all-False → empty catalog. Inverting the condition fixes both. (The catalog-side `get_pseudo_cls_catalog` has it right with `is_tomography`.)

Once inverted, note the assert above will also need `isinstance(..., (int, np.integer))` — bins from `np.unique` on a FITS column are `np.int64`, which fails `isinstance(x, int)`.

src/sp_validation/pseudo_cl.py
```diff
@@ -261,22 +694,137 @@ def get_pseudo_cls_catalog(
     )
     ell_eff = b.get_effective_ells()
 
-    factor = -1 if pol_factor else 1
+    is_tomography = tomo_bin_a != "all" and tomo_bin_b != "all"
+    if is_tomography:
+        mask_tomo_a = catalog[params["tomo_bin_col"]] == tomo_bin_a
```
cailmdaley, line 919: `params["tomo_bin_col"]` is never set anywhere (`get_params_rho_tau` doesn't know about it), so the catalog-path tomographic branch KeyErrors on first use. The map path reads the column name from `cc[ver]["shear"]["tomo_bin_ids"]` instead — worth picking one convention and threading it through both.

src/sp_validation/cosmo_val/core.py
```diff
@@ -222,6 +228,8 @@ def __init__(
         path_onecovariance=None,
         cosmo_params=None,
         blind=None,
+        compute_tomography=False,
+        force_run=False,
```
cailmdaley, line 232: (Anchoring here — the actual line, 222, is outside the diff.) `pol_factor` changed meaning from bool ("apply the −1 flip?") to the literal multiplicative factor (docstring: `int, default -1`) — but the default at line 222 still passes `True`, which multiplies e2 by 1, silently dropping the sign flip in the fiducial path. Should be `-1`; maybe also assert `pol_factor in (-1, 1)` to catch stray bools from old call sites.

src/sp_validation/cosmo_val/pseudo_cl.py
```diff
@@ -451,209 +449,359 @@ def calculate_pseudo_cl_g_ng_cov(self, gaussian_part="iNKA"):
...
-        shear_map_e2[unique_pix] += np.bincount(idx_rep, weights=e2 * w)
-        shear_map_e1[mask] /= n_gal_map[mask]
-        shear_map_e2[mask] /= n_gal_map[mask]
+    def calculate_pseudo_cl_map(self, ver, nside, out_path, tomo_bin_a, tomo_bin_b):
+        assert (tomo_bin_a == "all" and tomo_bin_b == "all") or (
+            isinstance(tomo_bin_a, int) and isinstance(tomo_bin_b, int)
+        ), "tomo_bin_a and tomo_bin_b must be either both 'all' or both integers."
 
-        shear_map = shear_map_e1 + 1j * shear_map_e2
+        params = get_params_rho_tau(self.cc[ver])
```
cailmdaley, line 615: Minor: the refactor dropped `survey=ver` here (and in `calculate_pseudo_cl_catalog`), which `calculate_pseudo_cl_eb_cov` still passes — survey-specific overrides in `get_params_rho_tau` silently revert to defaults. Intentional?

cailmdaley (commented):
Structure looks right — the bin-pair loop, the factored two-field NaMaster primitives, and the shared pixelization are exactly the shape this wants. Since you flagged it as not-ready, here's a pass aimed at getting it there: four concrete bugs inline below, all with a common root — the tomographic path (and the map path's noise-bias branch) has not yet been executed, and no test exercises a real bin pair (everything still runs `("all","all")`).

The one ask that closes all of this at once: a small end-to-end tomographic test — a tiny synthetic 2-bin catalog (or a #219 GLASS mock) through both estimator paths, checking that auto/cross spectra differ and E/B signs match the non-tomo path. It would have caught every finding below. Relatedly: this PR reads the bin column via `cc[ver]["shear"]["tomo_bin_ids"]` while #219 writes the literal column `TOM_BIN_ID` — neither PR adds the config entry wiring them together; worth adding it here so the GLASS validation loop actually connects.

— Fable, on behalf of Cail

src/sp_validation/cosmo_val/pseudo_cl.py
sachaguer: Ultimately, this should go away. This is hard-coding for some catalogue that is not relevant anymore and that should be removed in future PRs when the overall pipeline is stable.

martinkilbinger (approved):

sachaguer:
I pushed an extra update of the iNKA covariance estimation for tomography. I think many things changed in the code and this could be the minimum for this PR but not enough to close the linked issue yet. 

I will start testing the GLASS mock update jointly with the pseudo-Cl measurement.

cailmdaley:
I think I tiny unit test with a 2-bin setup would be nice to pin the current behavior and make sure there are no silly runtime errors :)

sachaguer:
I already have a set up, I will make the unit test when the more involved and relevant validation is done. Who can do more can do less

sachaguer:
I unintentionally merged the PR while bringing develop branch updates to the tomographic pseudo-cl branch. Apologies. I reverted the merge in `feature/sp_validation-extend-to-tomography`. 
I reopened a related PR #240

PR #225, opened 2026-07-03
Image-simulation m-bias: extracted core + snakemake orchestration

### LINKED ISSUE #224: Orchestrate image-sims end-to-end: m-bias core + snakemake
The sp_validation half of the image-simulation validation chain — extract the self-contained m/c-bias core from the divergent `image_sims_val` branch and write a fresh, fully-parameterized snakemake orchestration on the native `workflow/` framework that drives both containers (ShapePipe → merge → sp_validation calibrate → m-bias).

Sub-issue of the umbrella tracker UNIONS-WL/MultiBand_ImSim#1 (delivers its "generalize the downstream ShapePipe step" item). Pairs with CosmoStat/shapepipe#766.

Implemented in the PR that closes this issue.

— Claude, on behalf of Cail

### LINKED ISSUE #226: Additive-bias subtraction in calibration zeroes the signal on constant-shear sims (m = −1)
Found while running the image-sim m-bias chain end-to-end (#225): the m-bias came back at **m ≈ −1** (no shear response at all) even after the estimator itself was fixed. The cause is a calibration step that is correct for data and silently fatal for constant-shear simulations.

## Mechanism (exact, not just empirical)

`scripts/calibration/calibrate_comprehensive_cat.py` builds the science shear via `calibration.get_calibrated_m_c()`:

```python
# get_calibrated_quantities: correct ensemble metacal
g_corr = np.linalg.inv(R).dot(g_uncorr)          # R = ensemble 2x2 (shear + selection)

# get_calibrated_m_c: additive-bias removal — the problem for sims
c = np.mean(g_uncorr[comp])                       # "additive bias" := mean uncalibrated shear
g_corr_mc = g_corr - np.linalg.inv(R).dot(c)      # subtracted from every object
```

The multiplicative step is right. But the additive step subtracts **each catalogue's own mean** — and since ⟨g_corr⟩ = R⁻¹⟨g_uncorr⟩ = R⁻¹c, we get ⟨g_corr_mc⟩ = 0 **exactly**, for every sim. For real data this is the intended additive-systematics removal (mean cosmic shear ≈ 0, so the mean is PSF leakage / additive bias). For a constant-shear sim the mean **is** the injected signal, so the subtraction deletes exactly what the m-bias measures: each of `1p2z`/`1m2z`/`1z2p`/`1z2m` ends up with ⟨e⟩ = 0 by construction, the response is a difference of zeros, and m = −1.

## Evidence (grid_1, 1 tile, inverse-variance weighted, paired estimator)

| shear column | response 1+m₁ | response 1+m₂ |
|---|---|---|
| `e_uncal` (raw) | +0.84 | +0.92 |
| `e` = `g_corr_mc` (what the m-bias currently reads) | +0.20 | +0.02 ← signal gone |
| ensemble R⁻¹ on paired `e_uncal`, **no additive step** | +1.14 → **m₁ = +0.14 ± 0.10** | +1.27 → **m₂ = +0.27 ± 0.11** |

The last row is a hand-diagnostic using only the per-object `R_g**` columns (it omits the selection-response part of the ensemble R, a few-% term), so it's the cross-check that the pipeline is otherwise healthy — not the final number.

## Possible fixes — design call needed, since this touches the shared calibration path

1. **Sim-mode flag** (my leaning): an `additive_correction: true/false` knob on `calibrate_comprehensive_cat.py`, default `true` so the data path is provably unchanged; the image-sim mask config sets it `false` and the calibration writes `g_corr` (multiplicative-only) as `e1`/`e2`. Full ensemble R preserved; minimal diff.
2. **Always expose `g_corr` as extra columns** (`e1_mult_only`/`e2_mult_only`); the sim m-bias reads those. Non-invasive but adds columns for everyone.
3. **The m-bias estimator applies R itself** from `e_uncal` + `R_g**`. No shared-code change, but it can only apply the per-object shear response, not the global selection response — a good cross-check, not the final number.

Leaving the choice open — @-free ping via the PR thread. Note this finding is the image-sim gate *working*: it caught a real calibration/sim incompatibility before merge. Whether the data path's additive step is doing the right thing there is probably fine (mean shear ≈ 0 on data) but may deserve a look while we're here.

— Fable, on behalf of Cail

cailmdaley:
Closes #224
Closes #226

## Big picture

To trust a shear catalogue, we measure the bias of the pipeline that produced it: run the full measurement chain on simulated images with known injected shear, and compare what comes out to what went in. The result is the multiplicative and additive bias (m, c) — the numbers that calibrate, or validate, the survey's shear.

Four roles, three owners:

- **Sim campaign** (MultiBand_ImSim, external): sets of simulated tiles, identical except for the injected shear — one unsheared reference plus ± pairs in each shear component. Each set — a **branch** of the campaign — records its injected shear in `basic_info.txt`.
- **Measurement** (ShapePipe, [CosmoStat/shapepipe#766](https://github.com/CosmoStat/shapepipe/pull/766)): detection → PSF → shapes, run identically on every set.
- **Orchestration and estimation** (this PR): the Snakemake workflow that drives the whole chain, and the estimator that turns matched catalogues into m and c.

The measurement is only as good as the agreement between these components about what was injected, how things are named, and what columns mean. So the design rule throughout: **every such fact has exactly one home**, and everything else reads it or fails loudly.

## The chain

```
sim campaign (external)
  └─ basic_info.txt — the sims' own record of injected shear
        │ parsed, cross-checked
        ▼
im_manifest → manifest.yaml
  (branch → (g1,g2) map + derived amplitude; the single
   source of truth every downstream stage reads)
        ▼
Snakemake workflow — one container, committed SLURM profile
  ├─ im_pipeline    ShapePipe on each branch × tile
  ├─ im_merge       per-tile catalogues → final_cat_<branch>.hdf5
  ├─ im_extract     comprehensive ngmix catalogue
  ├─ im_calibrate   selection + metacal calibration
  └─ im_mbias       ImageSimMBias → m_bias_results.yaml + provenance
```

## What's here

**The campaign manifest.** A rule at the head of the DAG parses `g_cosmic` from every requested branch's `basic_info.txt`, derives the injected amplitude (cross-checked equal across the sheared branches, sign/axis checked against each branch's name), and writes `manifest.yaml` into the run root. The estimator takes the amplitude and branch map from the manifest alone; the manifest holds exactly the facts parsed from `basic_info.txt`.

**One estimator.** `ImageSimMBias` pair-matches the ±shear catalogues object-by-object so intrinsic shape noise cancels before averaging (a guard test locks the cancellation), and bootstraps the errors. Estimator variants are config points (`pair_match`, `w_cols`, `global_R_weight`) of this one class; a single run computes m/c for every weight scheme in `w_cols` (shared matching and bootstrap draws, first entry primary). It consumes the calibrated shear columns, so the headline quantity is **residual m**: the bias remaining after the chain's own metacal calibration. Fiducial estimator config: pair-matched, unweighted (#227), with additive-bias subtraction disabled (#226 — on constant-shear sims the subtraction removes the signal itself).

**Fail-fast configuration.** Science keys (`w_col`, `pair_match`, `match_radius_deg`, `n_bootstrap`, `bootstrap_seed`, `mask_config`) are required — no defaults in code or rule files. Operational keys default visibly in the workflow `config.yaml` and nowhere else. Run configs are schema-validated; unknown keys are errors. Tiles are given as a `tile_ids` list. The image-sim mask config is an explicit overlay on the base mask, so the sims-vs-data differences are declared.

**One container.** A single sp_validation image that can run ShapePipe — that's the one every rule uses.

**One run command.** `snakemake --profile candide …` with a committed SLURM profile (partition, node excludes, per-rule resources, `OMP_NUM_THREADS=1`). Snakemake owns all scheduling: it fans out across branch × tile jobs, and ShapePipe's internal multiprocessing is simply each job's thread count; the whole chain runs MPI-free.

**Provenance.** Beside each result: every resolved config value, the manifest hash, both repos' commits, the container tag. Derived tracking files are rebuilt each run.

## Validation

The current reference measurement (one tile, fiducial campaign, fresh from raw images) is **m1 = +0.0197 ± 0.0212, m2 = −0.0158 ± 0.0175, c1 = −0.0027 ± 0.0085, c2 = −0.0032 ± 0.0075** — consistent with zero at the percent level. The seeded bootstrap makes it reproducible, so it doubles as a regression check: refactors should leave it unchanged, and when a change moves it, the diff gets understood rather than waved through. The number itself will evolve as the analysis does. Synthetic-recovery tests (inject known m/c, assert recovery) cover the estimator itself.

Single-tile caveats, stated so the number isn't over-read: tile-to-tile variance is unmeasured at N=1; selection response is not yet accounted; the pair-matched estimator drops the few-percent tail of objects whose detection flips between branches.

## Before merge

- [ ] Reconciliation run at the companion estimator config point (pair-matched off, weighted) to confirm the one-estimator claim.

## Out of scope

Sim-generator changes (tracked as MultiBand_ImSim issues); the multi-tile campaign and selection-response study; cosmology-validation workflow rules (untouched).

— Fable, on behalf of Cail

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Development image with more bells and whistles
-FROM ghcr.io/cosmostat/shapepipe:develop
+FROM ghcr.io/cosmostat/shapepipe:im_sims
 
 RUN apt-get update -y --quiet --fix-missing && \
     apt-get dist-upgrade -y --quiet --fix-missing && \
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Development image with more bells and whistles
-FROM ghcr.io/cosmostat/shapepipe:image_sims-dev
+FROM ghcr.io/cosmostat/shapepipe:im_sims
 
 RUN apt-get update -y --quiet --fix-missing && \
     apt-get dist-upgrade -y --quiet --fix-missing && \
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Development image with more bells and whistles
-FROM ghcr.io/cosmostat/shapepipe:develop
+FROM ghcr.io/cosmostat/shapepipe:image_sims-dev
 
 RUN apt-get update -y --quiet --fix-missing && \
     apt-get dist-upgrade -y --quiet --fix-missing && \
--- a/config/calibration/mask_v1.X.4_im_sim.yaml
+++ b/config/calibration/mask_v1.X.4_im_sim.yaml
@@ -1,0 +1,62 @@
+# Config file for masking and calibration.
+# Standard cuts without coverage mask, type v1.X.4.
+
+# General parameters (can also given on command line)
+params:
+  input_path: shape_catalog_comprehensive_ngmix.hdf5
+  cmatrices: False
+  sky_regions: False
+  verbose: True
+
+# Masks
+## Using columns in 'dat' group (ShapePipe flags)
+dat:
+  # SExtractor flags
+  - col_name: FLAGS
+    label: SE FLAGS
+    kind: equal
+    value: 0
+
+  # Number of epochs
+  - col_name: N_EPOCH
+    label: r"$n_{\rm epoch}$"
+    kind: greater_equal
+    value: 2
+
+  # Magnitude range
+  - col_name: mag
+    label: mag range
+    kind: range
+    value: [15, 30]
+
+  # ngmix flags
+  - col_name: NGMIX_MOM_FAIL
+    label: "ngmix moments failure"
+    kind: equal
+    value: 0
+
+  # invalid PSF ellipticities
+  - col_name: NGMIX_ELL_PSFo_NOSHEAR_0
+    label: "bad PSF ellipticity comp 1"
+    kind: not_equal
+    value: -10
+  - col_name: NGMIX_ELL_PSFo_NOSHEAR_1
+    label: "bad PSF ellipticity comp 2"
+    kind: not_equal
+    value: -10
+
+# Metacal parameters
+metacal:
+  # Ellipticity dispersion
+  sigma_eps_prior: 0.34
+
+  # Signal-to-noise range
+  gal_snr_min: 10
+  gal_snr_max: 500
+
+  # Relative-size (hlr / hlr_psf) range
+  gal_rel_size_min: 0.5
+  gal_rel_size_max: 3
+
+  # Correct relative size for ellipticity?
+  gal_size_corr_ell: False
--- a/config/calibration/mask_v1.X.4_im_sim.yaml
+++ b/config/calibration/mask_v1.X.4_im_sim.yaml
@@ -60,3 +60,7 @@ metacal:
 
   # Correct relative size for ellipticity?
   gal_size_corr_ell: False
+
+  # Subtract additive bias (mean shear)? Use False for constant-shear
+  # image sims
+  additive_correction: False
--- a/config/calibration/mask_v1.X.4_im_sim.yaml
+++ b/config/calibration/mask_v1.X.4_im_sim.yaml
@@ -60,3 +60,7 @@ metacal:
 
   # Correct relative size for ellipticity?
   gal_size_corr_ell: False
+
+  # Subtract additive bias (mean shear)? Keep False for constant-shear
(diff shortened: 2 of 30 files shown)


cailmdaley:
Martin gave the green light in the corresponding shapepipe PR. i merged in my redesign and reran the whole pipeline, with no changes to the m-bias that comes out, so i am going to merge this and the corresponding shapepipe PR.

PR #230, opened 2026-07-06
Single shared mask-condition grammar (masks.py + cosmo_inference)

### LINKED ISSUE #181: Consolidate masking
Masking is done at several points:
- run_join.py:Mask with interface notebooks/create_binned_mask_comprehensive.py 
- cosmo_inference:masking.py (@LisaGoh)
Would be good to unify this.

cailmdaley:
Closes #181 (as rescoped in [this comment](https://github.com/CosmoStat/sp_validation/issues/181#issuecomment-4894536611)).

The object-selection `Mask` class and Lisa's cosmo_inference footprint builder parse the **same YAML mask-config format** through two independently-written condition evaluators whose `kind` vocabularies had diverged (`smaller_equal` vs `less_equal`; `greater`/`less` only in cosmo_inference). This PR gives them one grammar, leaving the object-space vs pixel-space split alone.

**Changes**
- `sp_validation.masks.apply_condition(array, kind, value)` — single evaluator, union of both `kind` vocabularies, with `smaller_equal` kept as an alias for `less_equal` so all existing configs (`config/calibration/mask_v1.X.{3,9,10,11}.yaml` are the only users of the old spelling) keep working unchanged.
- `Mask.apply()` delegates to it. Plain numpy instead of numexpr — single-comparison numexpr gains are marginal, and this drops a dependency (`numexpr` removed from `pyproject.toml`; it had no other consumers).
- `cosmo_inference/scripts/masking.py` imports the shared evaluator (its local copy deleted) and moves to `scripts/masking.py` — general use beyond inference, per Lisa's review. `--output-dir` is now required (the old default was relative to the script's location).
- New `src/sp_validation/tests/test_masks.py`: all 7 kinds, the alias, unknown-kind `ValueError`, and a `Mask`-vs-`apply_condition` equivalence check (10 tests, pass in the container).

No config files were touched; no behavior change for any existing YAML.

— Fable on behalf of Cail

https://claude.ai/code/session_01GPnbTHcheLscjVjnPdC6AH

### DIFF
--- a/cosmo_inference/scripts/masking.py
+++ b/cosmo_inference/scripts/masking.py
@@ -8,6 +8,8 @@ import healpy as hp
 import numpy as np
 import yaml
 
+from sp_validation.masks import apply_condition
+
 # -------------------------
 # Spatially-structured cuts: these define the survey footprint.
 # All other cuts (FLAGS, mag, SNR, shape measurement, PSF ellipticity,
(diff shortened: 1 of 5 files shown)


LisaGoh (approved):
Looks good, thank you for this! Just a suggestion, maybe `masking.py` should live in `sp_validation/scripts` instead, since it has a general use beyond purely for inference!

cailmdaley:
Done in 13b76c8 — moved to `scripts/masking.py`. Since the old output default was relative to the script's location, `--output-dir` is now required rather than silently pointing somewhere new; no other behavior change. (Left the hardcoded v1.4.5 input prefix as-is for now — happy to make it config-driven in a follow-up if useful.)

— Fable on behalf of Cail

PR #223, opened 2026-07-03
refactor(cosmology): move cosmology into cs_util.cosmo; delete local module

cailmdaley:
Companion to **CosmoStat/cs_util#76** and part of **CosmoStat/cs_util#75** — the two-PR consolidation of the collaboration's cosmology into `cs_util.cosmo` (the library home), per a conversation with Martin.

## What this does
Deletes `src/sp_validation/cosmology.py` (its theory-curve machinery now lives in `cs_util.cosmo` via #76) and repoints every consumer:

| Consumer | import now |
|---|---|
| `b_modes.py` | `from cs_util.cosmo import get_theo_xi` |
| `cosmo_val/core.py` | `from cs_util.cosmo import get_cosmo` |
| `cosmo_val/pseudo_cl.py` | `from cs_util.cosmo import get_theo_c_ell` |
| `papers/bmodes/scripts/precompute_pure_eb_chunk.py` | `from cs_util.cosmo import get_theo_xi` |
| `scratch/guerrini/namaster_utils.py` | `from cs_util.cosmo import get_cosmo, get_theo_c_ell` |

Also drops `"cosmology"` from `sp_validation.__all__`, updates two doc/comment references, and removes the now-duplicated `tests/test_cosmology.py` (moved to cs_util). **No call sites change** — the functions are byte-identical, only their home moved.

## Dependency
sp_validation now tracks cs_util's **`develop`** branch directly (git dependency) rather than a PyPI pin — cs_util releases are infrequent while the two repos iterate together:
```toml
cs_util @ git+https://github.com/CosmoStat/cs_util.git@develop
```
This repoint needs the new cosmology functions, which land via #76, so **this goes green once #76 merges into cs_util `develop`** — merge order is simply #76 → this. No release step, no version retarget.

Locally: all changed files compile, the repoint targets resolve in `cs_util.cosmo`, and no references to the old module remain.

— Claude on behalf of Cail

PR #236, opened 2026-07-08
Clean up cosmo_inference folder

### LINKED ISSUE #235: Clean up cosmo_inference folder
Currently, the `sp_validation/cosmo_inference` folder contents contains many scripts with functionalities that overlap with `cs_util` and `sp_validation`, as well as plotting notebooks that should be living in the `sp_validation/papers` directory. I propose to:

1. Clean up this directory by removing the obsolete scripts in the `cosmo_inference/scripts` subfolder
2. Clean up the `cosmo_inference/cosmocov_config` folder since we are no longer using Cosmocov
3. Move the notebooks to their relevant parent directories
4. Check that the structure of the folder and the scripts within adhere to the Snakemake framework in `sp_validation/workflow/rules/inference.smk`.

LisaGoh:
Addresses #235. So far I have:

1. Converted all the plotting notebooks in the folder to the relevant `papers/` subfolder and converted them to `.py` scripts to save space
2. Removed all but the `cosmosis_fitting.py` and `chain_postprocessing.py` scripts since they were overlapping/obsolete.
3. Cleaned up the `cosmosis_config` subfolder to only include the template ini files (and separated the ini files into `templates` and `ouput`, whereby the contents of `output` are only generated per user and not included in the repo to keep things lean)
4. Removed the `cosmocov_config` subfolder (in preparation for our switch to OneCovariance adoption instead)
5. Edited the `chain_postprocessing.py` script to use the `cs_util.cosmo` functionalities when computing the bestfit $\xi_\pm$ from the $C_\ell$'s

I have not yet run the `inference.smk` workflow myself, but per Claude the logic should not have been broken with these changes. I hope this helps!


cailmdaley:
Wow I love to see -17k lines :) and i think this is the first time we've seen the new ruff auto-check. Will review in a bit! I think we can leave inference.smk un-tested and fix any bugs when we start running chains on the GLASS mocks

cailmdaley:
hmm, @martinkilbinger could you please add Lisa as a maintainer to sp_validation so that the workflows run automatically and don't have to be approved manually each time? or you can make me an owner and i can do it :)

cosmo_inference/scripts/chain_postprocessing.py
```diff
@@ -281,15 +282,13 @@ def compute_best_fit_xi_from_cell(output_folder, root, best_fit_params, theta_ra
         output_folder + "{}/best_fit/shear_cl/bin_1_1.txt".format(root)
     )
 
-    import pyccl as ccl
-
-    cosmo = ccl.Cosmology(
+    cosmo = cs_cosmo.get_cosmo(
```
cailmdaley, line 285: `cs_util.cosmo.get_cosmo()` has different keyword names than `ccl.Cosmology` — its signature is `get_cosmo(Omega_b, Omega_m, h, sig8, ns, w0, wa, mnu, …, camb_params, extra_params)`, so `Omega_c=`, `n_s=`, `sigma8=`, and `extra_parameters=` will raise `TypeError` on the first call (and `matter_power_spectrum="camb"` isn't a valid CCL option either).

Since the chain parameters are already CAMB-format, the cleanest route is the wrapper's `camb_params=` path, roughly:

```python
cosmo = cs_cosmo.get_cosmo(
    camb_params={
        "H0": best_fit_params["h0"],
        "ombh2": best_fit_params["ombh2"],
        "omch2": best_fit_params["omch2"],
        "ns": best_fit_params["n_s"],
        "sigma8": best_fit_params["SIGMA_8"],
    },
    extra_params={
        "camb": {
            "halofit_version": "mead2020_feedback",
            "HMCode_logT_AGN": best_fit_params["logt_agn"],
        }
    },
)
```

(worth double-checking the exact `camb_params` keys against the wrapper).

cosmo_inference/scripts/chain_postprocessing.py
```diff
@@ -298,9 +297,7 @@ def compute_best_fit_xi_from_cell(output_folder, root, best_fit_params, theta_ra
         },
     )
 
-    theta_deg = np.rad2deg(theta_rad)
-    xi_p = ccl.correlation(cosmo, ell=ell, C_ell=shear_cl, theta=theta_deg, type="GG+")
-    xi_m = ccl.correlation(cosmo, ell=ell, C_ell=shear_cl, theta=theta_deg, type="GG-")
+    xi_p, xi_m = cs_cosmo.c_ell_to_xi(cosmo, theta_rad, ell, shear_cl)
```
cailmdaley: Units bug: `c_ell_to_xi` expects `theta` in **arcminutes** (it converts with `theta / 60` internally), but `theta_rad` here is in radians — so the correlation functions get evaluated at angles ~3438× too small. The previously-deleted code did the `np.rad2deg` conversion explicitly.

```python
xi_p, xi_m = cs_cosmo.c_ell_to_xi(cosmo, np.rad2deg(theta_rad) * 60, ell, shear_cl)
```

Keep saving `theta_rad` to `theta.txt` as you do below — the reader at L627 multiplies by `180/π·60`, i.e. it expects radians there.

cailmdaley (changes requested):
Thanks Lisa — this is a great cleanup, and the structure of it is exactly right: the `templates/` + `output/` split, the `.gitignore`, `inference.smk`, and `cosmosis_fitting.py` changes are all internally consistent, and −17k lines speaks for itself. 🎉

We did the light testing pass discussed in the thread (no full `inference.smk` run): `py_compile` on all 12 changed `.py` files passes, and we checked the new `cs_util.cosmo` calls against the cs_util source. That turned up **two real bugs in the new `chain_postprocessing.py` logic** (inline comments below) — both quick fixes — plus a few smaller things:

**Deletion fallout (worth fixing in this PR):**
- `workflow/rules/covariance.smk:256` still invokes the deleted `scripts/cosmocov_process.py`, and `papers/bmodes/scripts/run_cov_sweep.sh:68` calls the deleted `workflow/scripts/run_cosmocov_chain.sh`. Since CosmoCov is being retired for OneCovariance anyway, I'd say drop/update those callers here so nothing dangles. (We checked the `2pt_like_xi_sys.py` / `xi_sys_psf.py` references too — those resolve via `%(COSMOSIS_DIR)s` into the CosmoSIS install, so they're fine.)

**Dependency note (not this PR's fault, but it bites here first):**
- `get_cosmo` / `c_ell_to_xi` only exist on cs_util's development branch — released cs_util 0.2.1 (what the container has and what sp_validation pins) doesn't have them, so the import fails until a new cs_util release is cut. Worth coordinating with Martin on a release before merging, or at least noting it.

**Optional / non-blocking:**
- The converted `papers/realspace/*.py` scripts still carry some notebook artifacts — live `get_ipython()` / `display()` calls (`glass_mock_hist.py`, `get_chi2.py:562`, `unblinding_party_plots.py`), two `plt.savefig("/Plots/…")` paths in `contours.py` (632, 745) that would try to write to the filesystem root, and hardcoded `/home/guerrini` paths throughout. Since these are archival figure scripts that's fine — maybe just a one-line header note that they document the 2D analysis rather than being re-runnable as-is.
- `cosmosis_fitting.py`'s module docstring still says products land in `cosmosis_config/`; they now go to `output/`.

Once the two inline fixes are in, this looks good to us — and per the thread, we'll shake out any remaining `inference.smk` wiring when we start running chains on the GLASS mocks.

— Claude (Fable) on behalf of Cail

sachaguer:
Regarding `cs_util` an earlier PR on the develop branch pins the `cs_util` dependency to the develop branch. While still relevant to tag a version soonish, you could also merge the recent PR's of the develop branch in your branch.

Note also that the API of `cs_util.cosmo` has changed since this PR #223 and https://github.com/CosmoStat/cs_util/pull/79

LisaGoh:
Thank you for the review! I've updated this branch to match develop, so I'll keep referencing the develop branch of `cs_util` !

I've also added the `k_analysis.py` script that I used to calculate the k contributions to the correlation function for each theta.

cailmdaley:
Thanks Lisa! One more ruff pass and I think this is good to go

cailmdaley:
Sorry for all the pain with ruff @LisaGoh, these issues should be fixable with `ruff check --fix` and `ruff format`. I think Copilot was trying to manually fix the issue by grepping for whitespace etc... I'll make the developer/contributing instructions clearer on this. I think this can be merged!

PR #249, opened 2026-07-10
sacc_io converters: SACC → CosmoSIS 2pt-FITS (byte-verified) + SACC ↔ OneCovariance

### LINKED ISSUE #246: Converters: SACC->2pt-FITS + SACC<->OneCovariance
Supply the two glue layers that non-SACC consumers need: a SACC→2pt-FITS converter for the CosmoSIS inference chain, and n(z)→OneCovariance-input plus OneCovariance-output→SACC-covariance glue. Nothing here computes theory or blinds anything; it is theory-agnostic format work.

## Desired end state

- A converter turns a `{version}.sacc` analysis file into a 2pt-FITS file with the exact HDU layout the CosmoSIS chain (including the PSF ρ/τ likelihood modules in the cosmosis-standard-library fork) consumes. The chain runs against the converter's output and produces results identical to running against a directly-assembled file, with no pipeline change. The layout is an 11-HDU multi-extension FITS file; the executable definition of every HDU's columns and header cards is `cosmo_inference/scripts/cosmosis_fitting.py`.
- Two glue functions bridge OneCovariance: sp_validation n(z) tracers → OneCovariance's per-bin two-column text input (plus its config), and OneCovariance's flat output covariance → a SACC covariance block aligned to the analysis file's insertion order.

## Acceptance

- The 2pt-FITS converter is tested against a real reference file and a synthetic fixture (byte-compare where applicable).
- The OneCovariance glue round-trips through a fixture in both directions.

— Fable on behalf of Cail.

cailmdaley:
Closes #246. PR 3 of the PRD #241 series (base: #245, `sacc_io`).

## What & why

Extends **`sp_validation.sacc_io`** with converters between the SACC data product and external analysis-tool file formats. SACC stays the standard product; the current CosmoSIS inference chain runs unchanged behind it (PRD #241 §2: "converter first, native likelihood later"). One module now carries the whole SACC surface: format contract, guarded selection, and converters.

- **CosmoSIS 2pt-FITS**: `sacc_to_twopoint_fits(s, path, *, rho_stats_hdu=None, tau_stats_hdu=None, n_bins=1)` converts an analysis SACC into the 2pt-FITS read by `2pt_like` and the ρ/τ `2pt_like_xi_sys` fork. Single-bin v1: it raises on `n_bins != 1` and on any ξ pair beyond `(source_0, source_0)`. Tomographic emission lands with the tomographic round. RHO_STATS is copied verbatim from a sidecar FITS; the 2pt layout carries per-mode `varrho_*` variances the SACC does not store, and no consumer reads them.
- **OneCovariance**: writers for the combined-n(z) file and the `[redshift]` config stanza, plus a `covariance_list_*.dat` reader that produces dense blocks paired with SACC selectors for `assemble_covariance`. Coupled to OneCovariance only through its file formats, not as a dependency.

Every data and covariance selection goes through the guarded selection with `grid="reporting"` (loud on empty). Integration-grid points in the analysis SACC can never leak into the 2pt-FITS. (SACC covariances carry no tags of their own; blocks are addressed through data-point indices, which do — filtering the points filters the covariance with them.)

No public API outside `sacc_io` changes; no existing behavior is modified.

## Validation

Byte equality against the current writer (astropy writes FITS deterministically): build a reference with `cosmosis_fitting.py`'s own HDU-builders, build a SACC from the same inputs, convert, assert byte-for-byte identity.

- `test_sacc_io_twopoint.py` (CI): all three product shapes (plain-ξ; ξ+Cℓ; ξ+ρ/τ), the truncated-but-cross-correlated `[τ₀+; τ₂+]` covariance block, perturbation teeth, sidecar fail-fast, an encoded-covariance test that fails on any transposition, and byte-identity with integration-grid points present (regression pinning the grid-filtered covariance gather).
- `test_sacc_io_realdata.py` (candide-local `skipif`): byte-equal observed for real `SP_v1.4.6_leak_corr` and `glass_mock` products. No real data committed.
- `test_sacc_io_one_covariance.py`: reshaping, column selection, n(z) round-trip, fail-fast paths.

81 tests green, including the 4 real-data byte-compares on candide.

## Legacy writer

`cosmo_inference/scripts/cosmosis_fitting.py` remains the pipeline's 2pt-FITS writer for now; this converter is its byte-verified equivalent. PR 4 (#251) migrates the pipeline onto the SACC path and retires it. The swap happens downstream, not here.

— Claude (Fable, with Opus/Sonnet implementation agents) on behalf of Cail.

https://claude.ai/code/session_016Ko89uZF84ez6HEDcTWbJe


LisaGoh:
Thank you for this! After this is merged I can also adapt the OneCovariance snakemake workflow in #256 to include the sacc conversion

LisaGoh (approved):

sachaguer (approved):
Looks good to me. Hard to tell if everything is functionning at that stage but good place to start.

PR #250, opened 2026-07-10
Firecrown likelihood for the tomographic ξ± layout + CAMB↔CCL cross-check

### LINKED ISSUE #248: Firecrown likelihood for the tomographic ξ± layout + CAMB↔CCL cross-check
Row 5 of the PRD #241 §7 table — the theory engine the blinding box (PR 6) will call.

**Scope:**
- minimal firecrown likelihood for our tomographic ξ± layout: consumes the analysis SACC file; `build_likelihood()` + `compute_theory_vector()` is the whole surface Smokescreen uses; N-bin generic
- fiducial cosmology + model configuration (halofit version, IA model, neutrino treatment) as one named, clearly-configurable default set — the PRD §5 open question changes config values, not code
- CAMB↔CCL cross-check test: same cosmology, same n(z), ξ± compared on our θ grid against a stated tolerance, σ8-vs-A_s amplitude convention settled inside the test

Not in scope: Smokescreen wiring, secret handling, blinded-COSEBIs derivation (PR 6).

— Claude (Fable) on behalf of Cail.

cailmdaley:
Closes #248. PR 5 of the PRD #241 series (row 5 of §7); stacked on #245 (`sacc_io`).

## What

A minimal firecrown Gaussian ξ± likelihood (`sp_validation.blinding_likelihood`) — the theory engine PR 6's Smokescreen will call — plus a CAMB↔CCL cross-check test validating that the two theory stacks agree on our binning. No blinding wiring in this PR (that's PR 6).

## Why

Smokescreen shifts the data by `t(hidden) − t(fiducial)`, computed through firecrown + CCL, while inference runs CosmoSIS + CAMB. The shift only means what we intend if the stacks agree at fixed cosmology (PRD #241 §5). The likelihood is the only code that knows the data vector's exact layout (bin pairs, θ grid), so its `compute_theory_vector` is how theory gets aligned element-wise with the data.

## How it works

- Consumes the analysis SACC (`{version}.sacc`), extracting the coarse ξ± subset (past COSEBIs/pseudo-Cℓ/ρ/τ) with its aligned covariance — required, since firecrown models every point it reads and Smokescreen enforces `sacc.mean == likelihood data vector`.
- Generic over N tomographic bins (today's single-bin data = n=1); theory vectors realign to SACC insertion order via each statistic's `sacc_indices`, independent of statistic listing order.
- `TheoryConfig` is the single config surface: fiducial cosmology + model choices, defaults mirroring the CosmoSIS v1.4.6.3 `_A_cell` fiducial — S8=0.80, Ωm=0.30 (converted to CCL-native σ8/Ωc with the neutrino density correctly subtracted), Ωb=0.0469, h=0.70, nₛ=0.96, mν=0.06 eV, HMCode2020+feedback (logT_AGN=7.5) via CAMB, NLA wired but off by default (the shift nearly cancels IA; flipping to the inference stack's A=1.0 is one config value), normal-hierarchy neutrinos. The PRD's open fiducial choice changes config values, not code.
- Fine-grid ξ± shift is out of scope here — PR 6 computes it as a direct CCL theory difference.

## Validated

19 tests. Likelihood: construction from a `sacc_io` file, theory-vector shape/determinism/SACC-alignment (order-independence pinned against a genuinely reversed statistic order), 2-bin N-generality, config fail-fast. Cross-check: firecrown+CCL vs an independent CAMB-P(k)→CCL path agree to **0.21% (ξ+) / 0.10% (ξ−)** at fiducial and 0.25%/0.12% at a blind-sized offset (tolerances 0.5%/1.0%). The σ8-vs-A_s convention is settled inside the test — CAMB's A_s is iterated to share CCL's σ8 (a nominal A_s leaves σ8 ~3% off, which would blow the comparison to ~9–10%; the reconciliation is load-bearing and tested). The cross-check validates P(k)-recipe + amplitude consistency, not CCL's Limber+Hankel independently (both paths use it; documented in the module).

An independent adversarial review before opening drove the full attack list empirically and found **no correctness bugs** (verdict: open as-is). Highlights: the Ωm↔(σ8, Ωc, Ων) conversion verified to 5 decimals through firecrown's own CCLFactory path; **Smokescreen 1.5.6's actual loader driven end-to-end** — `ConcealDataVector` construction, `_verify_sacc_consistency`, concealing-factor calculation and application all succeed against this module, with the applied shift exactly matching the theory difference; extract-covariance alignment stress-tested on a 3-pair file with non-contiguous insertion order. One review note for PR 6: Smokescreen's loader never passes `theory_config` overrides — the module's defaults ARE the fiducial in the blinding path (overrides are reachable only from direct callers), and Smokescreen's `shifts_dict` values *replace* parameters rather than add (its documented-vs-actual gotcha).

Fast tests run in the default suite (~8 s); CCL/CAMB theory tests are `slow` with a fast smoke variant. Full fast suite green up to the two known pre-existing environmental failures (candide data paths from a worktree; snakemake absent from the local venv), both confirmed present at the base commit with this PR's files absent.

— Claude (Fable, with Opus implementation agents) on behalf of Cail.

https://claude.ai/code/session_016Ko89uZF84ez6HEDcTWbJe

### DIFF
--- a/src/sp_validation/blinding_likelihood.py
+++ b/src/sp_validation/blinding_likelihood.py
@@ -1,0 +1,364 @@
+"""Firecrown cosmic-shear likelihood — the blinding box's theory engine.
+
+:Name: blinding_likelihood.py
+
+:Description: A minimal `firecrown <https://firecrown.readthedocs.io>`_ Gaussian
+    likelihood over the tomographic real-space shear 2PCF (ξ±) written by
+    :mod:`sp_validation.sacc_io`. It exists for one job: give Smokescreen
+    (PR 6) a ``compute_theory_vector`` aligned element-for-element with the
+    measured ξ± data vector, so the Muir et al. blinding shift
+    ``d → d + t(hidden) − t(fiducial)`` can be evaluated. The sampling
+    likelihood stays with CosmoSIS; only the theory-vector predict step of this
+    module is ever called.
+
+    **Scope.** Only the *coarse analysis* ξ± subset (``grid='coarse'``). A
+    full analysis SACC file also carries pseudo-Cℓ, COSEBIs, pure-E/B and ρ/τ;
+    this likelihood extracts the coarse ξ± block (with its aligned covariance
+    sub-block) via :func:`sp_validation.sacc_io.extract` before handing it to
+    firecrown, because firecrown's ``read`` models every point it is given and
+    Smokescreen's ``_verify_sacc_consistency`` requires the SACC ``mean`` to
+    equal the likelihood's data vector element-for-element. The fine-grid ξ±
+    shift is computed in PR 6 as a direct CCL theory difference, not through
+    this likelihood.
+
+    **N tomographic bins.** Generic over any number of ``source_i`` NZ tracers:
+    it builds one firecrown ``WeakLensing`` source per bin and one ``TwoPoint``
+    (ξ+ and ξ−) statistic per unordered bin pair ``i ≤ j`` that the file
+    actually contains. Today's single-bin data is the ``n = 1`` case.
+
+    **Configuration surface.** :class:`TheoryConfig` is the single, explicit
+    home for the fiducial cosmology and model choices (nonlinear P(k) recipe,
+    intrinsic-alignment model, neutrino treatment). Its defaults mirror the
+    ``cosmo_inference`` CosmoSIS fiducial (v1.4.6.3 ``_A_cell`` pipeline) so the
+    two theory stacks agree by construction; PRD #241's open question about the
+    fiducial choice changes these *values*, not this code. Pass a
+    :class:`TheoryConfig` (or a mapping of overrides) through the firecrown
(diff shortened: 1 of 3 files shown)


cailmdaley:
Retiring this PR under the amended plan (see the plan-amendment comment on #241): the SACC/blinding stream now runs on the [UNIONS-WL/Smokescreen](https://github.com/UNIONS-WL/Smokescreen) fork, whose theory entry point is a plain `theory_fn` callable with a built-in CCL backend — the standalone firecrown likelihood this PR built is no longer a deliverable, and there is no successor PR. The CAMB↔CCL cross-check developed here survives: it moves into the blinding PR (#253) as a theory-consistency test on the same CCL backend that PR introduces (σ8-matched amplitude convention, matched `halofit_version` strings, 0.5%/1.0% ξ± tolerances).

This PR is to be closed without merging (close left to Cail).

— Fable on behalf of Cail.

PR #251, opened 2026-07-10
cosmo_val + Snakemake migration to the SACC writers (PRD #241 row 4)

### LINKED ISSUE #247: Migrate cosmo_val + Snakemake to the SACC writers
Make SACC the sole native output format of the cosmology-validation pipeline. Every two-point product (ξ± coarse and fine, pseudo-Cℓ, COSEBIs, pure-E/B, ρ/τ, n(z)) is written through the `sacc_io` builders into single-statistic SACC part files, and the Snakemake workflow assembles those parts into **one terminal SACC file per catalogue version**, `{version}.sacc`. No bespoke writer remains on the data-vector path.

## Desired end state

- Each `cosmo_val` statistic's write step is replaced by a call to the matching `sacc_io` builder, producing a part file that holds exactly one statistic plus the self-describing n(z) tracers it references. A producer writes a part with a single `save(s, path)` call; it never assembles a terminal file and never emits a bespoke format alongside.
- A dedicated Snakemake assembly rule reads the parts, concatenates them in canonical order, attaches the `BlockDiagonal` covariance (analysis blocks plus the dense per-pair fine block), and writes the single terminal `{version}.sacc`. The terminal file is a pure gather; anything that wants only one statistic targets the corresponding part rule, not a second terminal file.
- Part filenames are DAG internals: consumers bind parts through the workflow's path helpers, never by hardcoding names. `{version}.sacc` is the only stable name.
- **Blinding rides the DAG here.** Blinding is per part, at birth (#241 §4; CLIs and the assembly-time assertion are delivered by #252): a `blind-init` rule fires once per catalogue version, each blindable part rule (coarse ξ±, fine ξ±, pseudo-Cℓ) is followed by `blind-part` so only blinded parts persist on disk, COSEBIs/pure-E/B are derived from the blinded fine ξ± (born blinded), and the terminal assembly asserts `blind_commitment` is identical across parts. This issue authors the rules; #252 supplies the machinery.
- The bespoke `.txt`/custom-FITS/`.npz` writers on the data-vector path are deleted; nothing downstream reads them.

`sacc_io` is a fixed dependency and the single source of truth for every builder signature. Metadata keys required on every file: `catalogue_version`, `sp_validation_version`, `created`, `npatch`.

## Acceptance

The workflow produces `{version}.sacc` per version from parts, with only blinded blindable parts persisted and the commitment assertion exercised; the deleted bespoke writers have no remaining readers; the assembled file matches the `sacc_io` layout contract.

— Fable on behalf of Cail.

cailmdaley:
Closes #247. Row 4 of the PRD #241 §7 table: cosmo_val data products are **born as SACC**. Stacked on #245 (`sacc_io`, base branch `feat/sacc-2-sacc-io`), whose block-diagonal covariance and grid-tagged SACC layout are merged in here.

## What this does

- **Per-statistic SACC part writers** (`cosmo_val/sacc_writers.py` + mixin `*_to_sacc_part` methods). Each writer builds one statistic: reporting/integration-grid ξ± (TreeCorr → `xi_to_sacc`), pseudo-Cℓ EE/BB/EB with the NaMaster `BandpowerWindow`, COSEBIs at the fiducial scale cut, pure-E/B, and ρ/τ (τ theory covariance from the `CovTauTh` write-side).
- **One terminal file per version.** A new `assemble_sacc` rule loads the parts in canonical order, injects the externally-sourced covariance blocks (NaMaster FITS for pseudo-Cℓ; CosmoCov `--xi-cov` seam for ξ±), and writes the single terminal **`{version}.sacc`**. The terminal file carries the analysis vector only (ξ± reporting, pseudo-Cℓ, COSEBIs, pure-E/B, ρ/τ). The integration-grid ξ± is **not** gathered: it persists as its own per-part intermediate `{version}_xi_integration.sacc` (per the [#247 ruling](https://github.com/CosmoStat/sp_validation/issues/247#issuecomment-5033716753)), with Snakemake provenance covering traceability. This keeps the terminal file at tens-of-MB scale.
- **COSEBIs and pure-E/B consume the ξ± parts.** `cv_cosebis` / `cv_pure_eb` now derive their E-mode outputs from the born-as-SACC ξ± parts on disk, through two new values-only `b_modes` seams (`cosebis_from_xi`, `pure_eb_from_xi`) that call the same `cosmo_numba` kernels as the raw path. COSEBIs reads the integration part; pure-E/B reads the reporting + integration parts. The raw estimator run stays for the blind-invariant pieces (Bn, the jackknife covariance, the reporting-grid bin edges), and only the E-mode values are overridden from the part (`en_override` / `eb_override`), in both the SACC part and the diagnostic `.npz`. This makes born blinding automatic downstream: whatever the part carries flows into the E-modes with no blind-aware consumer code.
- **Integration ξ± is per-version, on a dedicated grid.** `xi_highres` runs per version (it was fiducial-only) as an in-container single-process TreeCorr job — at the config-driven 1000-bin grid the historical 10k-bin bare-host MPI path is unnecessary. The integration grid lives in its own `cosmo_val.integration` config block (`[0.08, 300]` @ 1000), so the one part serves both consumers: pure-E/B uses the full range (it must strictly contain the reporting grid), COSEBIs scale-cuts it up to 0.9. It is decoupled from `covariance.smk`'s own FIDUCIAL grid.
- **Assembly fails closed on unblinded data.** `assemble_sacc.py` no longer loads every part with `allow_unblinded=True`. The run type gates it: a `mock` run loads parts freely; a `data` run loads a part only when it carries the `concealed=True` stamp, else it raises. The run type comes from config (default `data`). This is the seam for PR #253's blind-at-birth — once each data part is concealed at write time, the `data` run assembles with `allow_unblinded=False` untouched.
- **Covariance is block-diagonal.** `assemble_covariance` (from the base) builds a validated `BlockDiagonalCovariance` (one FITS table per block), not a dense `FullCovariance`. Ordering is validated (contiguous, tiling, square). The assembled `.dense` is unchanged; cross-blocks are zero.

## Branch hygiene

A stale "Rebuild PR4" base re-added the `cosmo_inference/` analysis tree that develop deleted in #236 (the develop merge kept it, since a delete-vs-readd does not conflict). This PR removes those 80 resurrected files (notebooks, the cosmosis_config `.ini` set, scripts, `get_chi2` notebooks, pipeline shells). The diff versus develop drops from 150 changed / 87 added to **72 changed / 7 added** — roughly the SACC migration itself.

## Design notes for reviewers

- **`xi` and `xi_highres` stay two rules.** A single binning-agnostic rule was considered but not forced: the two drivers genuinely diverge (`CosmologyValidation.calculate_2pcf` with a born-cov-less reporting part vs. standalone TreeCorr with a `DiagonalCovariance` from `varxip`/`varxim`), so unifying them would carry its own validation burden for no gain here.
- **Placeholder covariance is explicit opt-in.** With no real ξ-covariance wired, assembly fails loudly by default. `cosmo_val.allow_placeholder_cov: true` exists for dry-runs and tests only. The placeholder is flagged and never a science covariance.
- **blind=A in the analysis file.** The pseudo-Cℓ part feeding `{version}.sacc` is the tagged legacy inference product (`blind=A_powspace_nbins=32`). The measured spectra are blind-independent and the iNKA covariance is analytic, so the tag is provenance-naming, not wrong data.
- Note the pre-existing `blind` wildcard on the `pseudo_cl` rules is the glass-mock A/B/C variant (three mock catalogues), **not** Smokescreen blinding. A rename is avoided because that name is baked into on-disk filenames we do not own; prominent comments now mark the distinction.
- `inference.smk` (dormant) still names pre-SACC inputs. Migration to consuming `{version}.sacc` is the native-sacc PR (row 7).

## Blinding integration

Smokescreen blinding is a separate axis, tracked by #241/#247. The blind-init and blind-part DAG rules land in **PR #253**, which stacks on this branch. This PR is version-agnostic and blind-free; it leaves the fail-closed loading seam ready, and the part-consuming design means born blinding needs no blind-aware consumer code on the stacked branch.

## Tests

The migration, `sacc_io`, and cosmo_val suites pass (`-m "not slow"`): 219 passed, 1 skipped. The `assemble_sacc` and paper B-mode DAGs resolve in a dry-run. One candide-path guard fails only on nodes with catalogue-mount gaps (environmental).

— Claude (opus) on behalf of Cail.

https://claude.ai/code/session_01EaL7prmKUHwJQcDyW3LoxD


cailmdaley:
@sachaguer and @LisaGoh i believe this is ready for review! this was a messy PR since it touches all the snakemake rules and their scripts. it's been hard to keep the scope focused, there is probably some lingering slop to be cut out as we critique.

LisaGoh:
Sorry for the late reply! I wasn't able to review every single file as this is a pretty large PR, but the overall logic seems sound. Happy to merge this first to keep the restructuring going!

LisaGoh (approved):

LisaGoh:
Another aside: I'll put aside the development of the tomographic extension for now, since it's getting rather messy with these two efforts (sacc integration + tomographic extension) running in parallel. Let's fix the sacc io format first, then I'll update the PRs on the tomographic xi_pm and covmat calculations to incorporate this format so effort isn't wasted; what do you think?

sachaguer (approved):
sorry for the long overdue review.

I think the spirit of the PR is good and it seems well propagated in the scripts. I could not review everything as it is a very large and dense PR but I think it is good to merge and the ultimate test will be whether we find any bugs using it.

PR #285, opened 2026-07-17
Calibration fail-fast, pipeline_config_dir, jobs cap, response docs (imsims A/B campaign)

### LINKED ISSUE #284: Calibration robustness + image-sims workflow fixes from the A/B campaign
Four small items surfaced by the shape-measurement A/B characterization campaign (shapepipe#813):

- **Empty metacal selection dies with a bare pandas error** ("cannot set a frame with no defined index") deep in `get_w_des`, when the real condition — zero objects survive selection because upstream shape measurement failed wholesale — is knowable at `metacal` construction. Needs a fail-fast guard that names the actual problem.
- **`pipeline_config_dir`** — the image-sims workflow needs a per-run override for the ShapePipe config tree (the surface that owns the ngmix ini), so config-only A/B arms don't have to fork the shared worktree. Proven in campaign use; unset ⇒ byte-exact prior behaviour.
- **Documented invocation is broken under snakemake ≥ 9**: the Snakefile docstring says no `-j` needed, but the candide profile carries no `jobs:` key and snakemake 9.16.3 hard-requires one.
- **Global-response calibration is undocumented**: `gal_metacal.R` is a run-averaged 2×2 matrix applied to every object; per-object responses are computed but only ever consumed as means. Worth one explicit docstring.

— Claude (Fable), on behalf of Cail.

cailmdaley:
Closes #284

Four small fixes from the A/B characterization campaign (shapepipe#813), one commit each:

- **`pipeline_config_dir`** (cherry-pick from the campaign branch): optional run-config key pointing the ShapePipe pipeline stage at a per-run config tree; unset ⇒ byte-exact prior behaviour. This is what let every config-only A/B arm run without forking the shared worktree.
- **Fail fast on empty metacal selection**: `metacal._masking_gal` now raises a named error ("0 objects survive the metacal selection … upstream shape measurement likely failed wholesale") instead of letting a 0-row frame die three layers down in pandas.
- **snakemake ≥ 9 jobs cap**: docstring corrected *and* the candide profile gains `jobs: 100`, so the documented profile-only invocation works again (9.16.3 hard-requires a job cap).
- **Document global-response calibration**: one Notes block on `get_calibrated_quantities` stating that a single run-averaged R is applied to every object (per-object responses exist but are only consumed as means).

Reviewer call intentionally left open: none here — the debatable piece (RuntimeError on wholesale failure) lives in the shapepipe sibling PR.

— Claude (Fable), on behalf of Cail.

https://claude.ai/code/session_019dA9sas4vvNBcNKRmi973L

### DIFF
--- a/src/sp_validation/calibration.py
+++ b/src/sp_validation/calibration.py
@@ -40,6 +40,16 @@ def get_calibrated_quantities(gal_metacal):
         weights
     mask : array of bool
         mask to indicate valid objects in "no-shear" sample
+
+    Notes
+    -----
+    Calibration uses a single global response: ``gal_metacal.R`` is the
+    2x2 shear+selection response matrix averaged over the whole run, and
+    the same matrix is applied to every object. Per-object responses
+    (``R_g11`` ... ``R_g22``) are computed upstream but are only ever
+    consumed as catalogue-wide (or, in ``get_w_des`` /
+    ``get_quantities_binned``, per-bin) means -- no object is calibrated
+    with its own response.
     """
     # mask for 'no shear' images
     mask = gal_metacal.mask_dict["ns"]
--- a/src/sp_validation/calibration.py
+++ b/src/sp_validation/calibration.py
@@ -976,6 +976,14 @@ class metacal:
 
             self.mask_dict[name] = ind_masked
 
+        if len(self.mask_dict["ns"]) == 0:
+            raise ValueError(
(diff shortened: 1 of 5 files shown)


cailmdaley:
Closing — superseded by CosmoStat/shapepipe#891 (one workflow for data and image sims). The calibration fail-fast and the snakemake jobs cap from here will come back on their own once that work starts. — Fable on behalf of Cail

PR #291, opened 2026-07-20
Bug in iNKA covariance

### LINKED ISSUE #290: Bug: iNKA analytic covariance does not input decoupled noise bias
In the computation of the iNKA covariance, a fiducial Cell is extracted using `cs_util` and a noise bias is computed from the catalog. Two branches are possible:
- `randoms`: Generate multiple Gaussian noise only maps from the catalog and averages the decoupled cells of many realisations. Computes the decouples spectra as is.
- `analytical`: Uses an analytical description from Nicola et al. (2019) to get the noise bias. *The analytical description returns the coupled noise bias. The decoupling is missing in the current implementation.*

This bug was caught during the validation of #282 #286 .

sachaguer:
Related to Issue #290 , this PR fixes a bug in the iNKA covariance. The analytical based noise bias computation was missing a decoupling step leading to a misestimated noise bias and likely larger error bars. The updated functions is running for cross-validation with Gaussian simulations from #286 .

Ready to be merge as soon as the cross-validation is performed. In the meantime, design and runtime review is welcomed.

### DIFF
--- a/src/sp_validation/cosmo_val/pseudo_cl.py
+++ b/src/sp_validation/cosmo_val/pseudo_cl.py
@@ -224,7 +224,7 @@ class PseudoClMixin:
                     noise_bias_cl = self.get_noise_bias(params, nside, cat_gal_)
 
                     # If using the analytical noise bias, we need to decouple it
-                    if self.noise_bias_method == "analytical":
+                    if self.noise_bias_method == "analytic":
                         wsp = wsp_dict[f"W{bin_key1}xW{bin_key2}"]
                         noise_bias_cl = wsp.decouple_cell(noise_bias_cl)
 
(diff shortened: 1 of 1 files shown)


sachaguer:
@cailmdaley, following #286, this PR is ready for a review.

cailmdaley:
The fix is correct. I traced the decoupling end to end. The coupled analytic noise from Nicola et al. is now decoupled with the correct workspace. Coupled and decoupled quantities are used consistently downstream. Cross-bins correctly get zero noise bias. Approve.

One request, on a pre-existing issue this fix now touches: the signal and noise grids are misaligned by one multipole. `get_fiducial_cl` builds the signal on `arange(1, lmax+1)`, so index i holds C(ℓ=i+1). The NaMaster output is 0-indexed, so index i holds N(ℓ=i). The sum `fiducial_cl + noise_bias_cl` therefore adds N(ℓ) to C(ℓ+1). The effect is small while the noise is flat, but please unify the two grids on the NaMaster convention (`arange(0, lmax)`, ℓ=0,1 zeroed) so the fix rests on aligned inputs.

— Claude on behalf of Cail

sachaguer:
The best I found to fix this is to add a zero monopole and drop the last value of the fiducial.
CAMB crashes when computing a theory at ell=0 so it seems to be the best solution to me. Happy to do something else if someone has a better idea.

PR #295, opened 2026-07-22
Tomographic rho-/tau-statistics

### LINKED ISSUE #216: Tomographic rho-/tau-statistics
**Goal**

Upgrade the existing rho/tau-statistics measurement pipeline to tomography

**To do**

- Code already exists within Euclid's FDQA and could be easily incorporated.
- Additional tests such as tangential shear around stars is related and could also imported within the scope of this issue.

sachaguer:
This PR upgrades the existing script to compute rho-/tau-statistics to iterate over tomographic bins. Only the tau-statistics change with the bin (the sample of star remains fixed).

The format to call tomographic run is similar to the implementation for pseudo-cl #221. Plotting functions have been upgraded to handle it as well.

The code has run on the GLASS mock, see the plot below for a validation example. The PSF error parameters are consistent with zero at the 2sigma level across the bins.

image


src/sp_validation/cosmo_val/core.py
```diff
@@ -483,11 +483,20 @@ def results_objectwise(self):
...
+        npatch=None,
+    ):
         cfg = treecorr_config or self.treecorr_config
         patches = npatch or self.npatch
+        tomo_bin_a_str = f"tomo_bin_{tomo_bin_a}"
+        tomo_bin_b_str = f"_tomo_bin_{tomo_bin_b}" if tomo_bin_b is not None else ""
         return (
-            f"{version}_minsep={cfg['min_sep']}"
+            f"{version}_{tomo_bin_a_str}{tomo_bin_b_str}_minsep={cfg['min_sep']}"
             f"_maxsep={cfg['max_sep']}"
```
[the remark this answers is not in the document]
sachaguer: This will have to be checked whether new outputs are read differently or if the call to open the file relies on the basename function as well.

cosmo_val/cat_config.yaml
```diff
@@ -1280,7 +1280,7 @@ GLASS_mock_validation:
     star_flag: FLAG_STAR_HSM
     star_size: SIGMA_STAR_HSM
     hdu: 1
-    path: unions_shapepipe_psf_2024_v1.6.a.fits
+    path: ../../../../n17data/UNIONS/WL/v1.6.x/unions_shapepipe_psf_2024_v1.6.a.fits
```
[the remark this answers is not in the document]
sachaguer: Not a problem here, this is just for testing.

sachaguer:
I updated the plotting script. In particular, I wrote a wrapper to plot any 2PCF like tomographic plot that requires to be fed a function to plot the actual thing on the axes. This seems quite flexible to me and could be used outside of the pipeline to produce plots from any other output. Ultimately could be useful to plot the 2PCF data vectors (together with the leakage) (see #296 )

image

image

sachaguer:
Updates also handle the non tomographic case with a call to the same function

image

sachaguer:
@cailmdaley @LisaGoh Ready for a review and probably relevant for #296 

Also related to https://github.com/CosmoStat/shear_psf_leakage/pull/35

src/sp_validation/cosmo_val/psf_systematics.py
```diff
@@ -341,35 +376,548 @@ def set_params_leakage_object(self, ver):
...
-                        f"Skipping computation, reading {output_path_ab} and "
-                        f"{output_path_aa} instead"
+    def set_psf_parameter_nsamples(self, nsamples):
+        self.psf_error_nsamples = nsamples
+
+    def set_psf_parameter_nwalkers(self, nwalkers):
+        self.psf_error_nwalkers = nwalkers
+
+    def get_samples(self, version, params, tomo_bin_id, track_result=False):
+        npatch = params["patch_number"] if self.cov_estimate_method == "jk" else None
```
cailmdaley, line 392: `set_params_rho_tau` no longer sets `patch_number`, but this line still reads it when `cov_estimate_method="jk"` → KeyError. Since the class already carries `self.npatch`, maybe use that directly?

src/sp_validation/cosmo_val/psf_systematics.py
```diff
@@ -137,150 +54,268 @@ def psf_fitter(self):
             )
         return self._psf_fitter
 
-    def calculate_rho_tau_fits(self):
+    @property
+    def rho_tau_fits(self):
```
cailmdaley, line 58: `calculate_rho_tau_fits` defaults `tomography=True`, and this property never calls it with `False` — so `"tomo_bin_all"` is never populated, but `plot_rho_tau_fits()` (default non-tomo) reads it unconditionally → KeyError on every call. The same pattern as your other properties (`tomography=False` first, then `True` under `compute_tomography`) would restore the non-tomo path. `xi_psf_sys` below has the same issue, plus it will raise on a genuinely non-tomographic catalogue.

src/sp_validation/rho_tau.py
```diff
@@ -66,43 +52,60 @@ def get_rho_tau_w_cov(
...
-    npatch=None,
+    ncov=100,
+    compute_minus=True,
+    **kwargs,
 ):
     """Compute rho/tau statistics and, if requested, their covariance."""
     if method == "th":
-        nbin_ang, nbin_rad = 100, 200
+        nbin_ang, nbin_rad = kwargs.get("nbin_ang", 100), kwargs.get("nbin_rad", 200)
+        compute_minus = kwargs.get("compute_minus", True)
```
cailmdaley, line 67: minor: `compute_minus` is a named parameter, so it never lands in `**kwargs` — this `kwargs.get("compute_minus", True)` always returns True and passing `False` is silently ignored. Just use the argument directly.

cailmdaley (changes requested):
The estimator plumbing (per-bin masks, basename, base_rho/base_tau split) reads internally consistent, and the GLASS validation is convincing. Two paths look broken where the new defaults meet the old accessors — both easy fixes — plus one dead switch.

One non-blocking aside: effective number density / survey area are now computed independently in `shear_psf_leakage` (for the tau theory covariance) and in `sp_validation` (for the OneCovariance config, #298). Nothing wrong here, but we opened CosmoStat/cs_util#81 to converge on one canonical implementation so the two covariances can't silently disagree on shape noise — opinions welcome there.

— Claude, on behalf of Cail

sachaguer:
Bugs identified are fixed in the last commit.

Agreed for the homogeneisation of the number density, area, and shape noise computations. Should we keep it for a future PR as the branches for tho-/tau- and OneCovariance are modified indendently right now?

Also, I left the scale dependant leakage and the objectwise leakage aside for now but this could be part of this PR if I find some time to do it.

cailmdaley:
Let's leave the survey statistics unification for a future PR! Up to you on the leakage

cailmdaley (approved):

LisaGoh:
Thank you for this, looks good to me too! As I was suggesting in #296 perhaps after merging this we can move the `plot_2pcf_tomography` function to `real_space.py` and all the real space plotting goes there, while the x_y_functions relavant to the psf systematics stay here

LisaGoh (approved):

PR #298, opened 2026-07-23
Update routines to call OneCovariance for tomography

### LINKED ISSUE #283: OneCovariance with tomography
This issue aims to update the existing code to compute OneCovariance harmonic space covariance with tomography.

OneCovariance already allows for such computation. What needs to be implemented is the interface between our catalogue and OneCovariance.

sachaguer:
This PR should be quite small. OneCovariance already handles tomographic run. These commits are aimed at modifying the functions setting up the config file to compute a tomographic Cell covariance with OneCovariance. 

In particular, it requires the computation of shape noise and galaxy density per tomographic bin.

This PR is related to the issue #283 .

### DIFF
--- a/src/sp_validation/cosmo_val/catalog_characterization.py
+++ b/src/sp_validation/cosmo_val/catalog_characterization.py
@@ -150,7 +150,9 @@ class CatalogCharacterizationMixin:
     @property
     def n_eff_gal(self):
         if not hasattr(self, "_n_eff_gal"):
-            self.calculate_n_eff_gal()
+            self.calculate_n_eff_gal(tomography=False)
+            if self.compute_tomography:
+                self.calculate_n_eff_gal(tomography=True)
         return self._n_eff_gal
 
     @property
--- a/src/sp_validation/cosmo_val/catalog_characterization.py
+++ b/src/sp_validation/cosmo_val/catalog_characterization.py
@@ -156,7 +156,9 @@ class CatalogCharacterizationMixin:
     @property
     def ellipticity_dispersion(self):
         if not hasattr(self, "_ellipticity_dispersion"):
-            self.calculate_ellipticity_dispersion()
+            self.calculate_ellipticity_dispersion(tomography=False)
+            if self.compute_tomography:
+                self.calculate_ellipticity_dispersion(tomography=True)
         return self._ellipticity_dispersion
 
     def _get_binned_catalog_mask(self, ver):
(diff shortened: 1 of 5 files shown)


src/sp_validation/cosmo_val/pseudo_cl.py
```diff
@@ -507,18 +538,23 @@ def _modify_onecov_config(
...
         )
         gaussian_one_cov = cov_from_one_covariance(cov_one_cov, gaussian=True)
         all_one_cov = cov_from_one_covariance(cov_one_cov, gaussian=False)
 
-        self._pseudo_cls_onecov[ver] = {
-            "gaussian_cov": gaussian_one_cov,
-            "all_cov": all_one_cov,
-        }
+        key_to_update = "tomo" if tomography else "non_tomo"
+        self._pseudo_cls_onecov[ver].update(
```
cailmdaley: `self._pseudo_cls_onecov[ver]` is never initialized before `.update()` — the old code assigned directly, so this raises KeyError on the first call for each `ver` (non-tomographic path included). `self._pseudo_cls_onecov.setdefault(ver, {}).update(...)` fixes it.

src/sp_validation/cosmo_val/catalog_characterization.py
```diff
@@ -156,7 +156,9 @@ def n_eff_gal(self):
     @property
     def ellipticity_dispersion(self):
         if not hasattr(self, "_ellipticity_dispersion"):
-            self.calculate_ellipticity_dispersion()
+            self.calculate_ellipticity_dispersion(tomography=False)
+            if self.compute_tomography:
```
cailmdaley, line 162: `ellipticity_dispersion` got the `compute_tomography` trigger, but its twin `n_eff_gal` didn't — so `_modify_onecov_config(tomography=True)` will look up `n_eff_gal[ver]["tomo_bin_i"]` keys that were never computed. Mirroring the same two-line pattern onto `n_eff_gal` should do it.

cailmdaley (changes requested):
Nice and compact — the config plumbing does what it says. Two small initialization gaps will KeyError on first use; both are one-line fixes.

— Claude, on behalf of Cail

cailmdaley:
oops sorry Sacha! i meant to review https://github.com/CosmoStat/shear_psf_leakage/pull/35 instead of this PR, which i see may be in progress and which you hadn't requested my review on yet.

sachaguer:
The OneCovariance wrapper now runs in tomographic case (see figure below).

image

@cailmdaley this is ready for a review. A few things to note during the review:
- I checked, but worth keeping an eye on it during the review, that the ordering of the blocks in the covariance match between the iNKA merge I wrote and the way OneCovariance saves the data natively. I looked at the diagonal of the two covariances and there is a decent match so I think this is correctly done.
- I met some hurdles on the way related to package versioning. We currently install numpy==2.5.0 (the most recent version?) using the `pyproject.yoml`. But OneCovariance crashes with `numpy>2`. I opened an issue https://github.com/rreischke/OneCovariance/issues/34. I updated the code to run on my own fork and OneCovariance is in my knowledge not part of the package yet. We might want to add it reading my fork for now but potentially from Robert Reishke's repository if he solves the issue.
- I also made some updates to the iNKA part.

cailmdaley:
happy rentrée Sacha! this looks good. review incoming

cailmdaley:
i'm happy merging this as-is. we can also put your OneCovariance fork in the `pyproject.toml` so no one else faces the same hurdles, thanks for making that fix

